Issue #2, opened 2020-09-22
Use external cli argument parser


qarmin:
For now arguments are parsed manually, but some people find hard to read my code, so maybe better idea is to use some lightweight argument parser.

Some of possible libraries:
- https://crates.io/crates/gumdrop
- https://crates.io/crates/structopt

MeirKlemp:
I would like to work on it.
Can I be assigned to it?

sseering:
While doing this one should also change 

`--v / --version` to `-v / --version`

and 

`--h / --help` to `-h / --help`.

The double dashes are kinda a bug/misfeature that is the result of the current implementation of the argv parser.

qarmin:
Fixed by #37

PR #37, opened 2020-10-03
Changed to StructOpt for CLI argument parser.

MeirKlemp:
I finished the task of this issue #2 with the crate StructOpt.

## Some changes from the previous parser
- I used subcommands: `$ czkawka --d` is now `$ czkawka dup` and so on.
- I changed some option short names: `-E` for excluded items instead of `-k` to be closer to the option name (you can change it back if you want to).
- I removed the -f option because you can achieve the same result with the > operator.
  **example**: `$ czkawka dup -d ~ > results.txt` it will create the file results.txt and put stdout buffer in it.
  **note**: In order to print the messages into the console you can use the stderr with `eprintln!` macro.


qarmin:
Awesome work.

I have only some thoughts:
- When I run `czkawka_cli` or `czkawka_cli -h` I wasn't sure what I need to do.
I think that there should be information like `use czkawka_cli <SUBCOMMAND> -h to get more info about specific tool` or something equivalent to it.
- Running `czkawka_cli big` prints a lot of duplicated infos 
```
BiggestFiles { directories: Directories { directories: ["/home/rafal", "/home"] }, excluded_directories: ExludedDirectories { excluded_directories: [] }, excluded_items: ExludedItems { excluded_items: [] }, allowed_extensions: AllowedExtensions { allowed_extensions: [] }, number_of_files: 50, not_recursive: NotRecursive { not_recursive: false } }
```
Three times each word like `directories`, `not_recursive` etc. is written but I think that one should be enough.
- in help `czkawka_cli <SUBCOMMAND>` is little misleading, because it suggest that `czkawka_cli big` should be enough to run app. Maybe something like `czkawka_cli <SUBCOMMAND> <SCFLAGS> <SCOPTIONS>` will be more appropriate?
- `-d directories` is only required parameter, but It looks exactly same as other optional parameters 
- In earlier version there were "real world" examples like `czkawka --d -i "/home/rafal/,/home/szczekacz" -e "/home/rafal/Pulpit,/home/rafal/Obrazy" -s 25 -x "7z,rar,IMAGE" -l "hashmb" -f "results.txt" -delete "aeo"` which I think is very helpful for users

About removing builtin saving to file, I'm opposite to this, because this will prevent users of Windows/Mac(even if now this OS are not supported officially) easily usage of this feature.
Also I was thinking about `quiet` feature, so users can easily select if they want results printed to CLI or saving to file or both without needing use `>` or `2>&1 | tee`.

MeirKlemp:
Thanks for reviewing.

- The help messages are built-in, they should be modifiable but right now I don't know how to do it. (also for examples)
- In the printed debug message you see a lot of duplications because I used the flatten technique so I can use `Duplicates` in all subcommands without copying the whole information of the option. I can change it to be `BiggestFiles { d: D { directories: ["~"] }, ...`.
- I just found out how to make `directories` required by structopt.
- Windows/macOS/Linux has the operator `>`. I think it is better to add `-f` but also make it work with `>` because a lot of cli users are using this operator.
- I don't exactly understand how the option `quiet` will work.

MeirKlemp:
I can see now that it is possible to change the header template with [clap](https://docs.rs/clap/2.33.3/clap/struct.App.html#method.template) (StructOpt is a wrapper of clap)

qarmin:
`quiet` option just will prevent from printing results to screen
Code 
```
df.print_results();
df.get_text_messages().print_messages();
```
after changes will be
```
if !quiet_mode{
df.print_results();
df.get_text_messages().print_messages();
}
```
but that goes beyond this PR

MeirKlemp:
Please review my last commit.

qarmin:
Can you rebase this PR?
In latest commit I added to CI automatically export of artifacts, so anyone will able to check this PR without needing manually recompile app.

MeirKlemp:
I am sorry, but I don't know how to rebase and why we should rebase now...

czkawka_cli/src/commands.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub excluded_items: Vec<PathBuf>,
+}
+
+#[derive(Debug, StructOpt)]
+pub struct AllowedExtensions {
+    #[structopt(
+        short = "x",
+        long,
+        help = "Allowed file extension(s)",
+        long_help = "List of checked files with provided extension(s). There are also helpful macros which allow to easy use a typcal extensions like:\nIMAGE(\"jpg,kra,gif,png,bmp,tiff,webp,hdr,svg\"),\nTEXT(\"txt,doc,docx,odt,rtf\"),\nVIDEO(\"mp4,flv,mkv,webm,vob,ogv,gifv,avi,mov,wmv,mpg,m4v,m4p,mpeg,3gp\") or\nMUSIC(\"mp3,flac,ogg,tta,wma,webm\")"
```
qarmin: Here probably should be 1 more empty line - 
Zrzut ekranu z 2020-10-04 20-36-52

czkawka_cli/src/commands.rs
```diff
@@ -0,0 +1,192 @@
...
+USAGE:
+    {usage} [SCFLAGS] [SCOPTIONS]
+
+FLAGS:
+{flags}
+
+SUBCOMMANDS:
+{subcommands}
+
+    use {usage} -h to get more info about a specific tool
```
qarmin: This looks in code that
```
 use czkawka_cli <SUBCOMMAND> -h to get more info about a specific tool
```
but quotes, brackets or something similar should be added to make text easier to read
```
use "czkawka_cli <SUBCOMMAND> -h" to get more info about a specific tool
use [czkawka_cli <SUBCOMMAND> -h] to get more info about a specific tool
use `czkawka_cli <SUBCOMMAND> -h` to get more info about a specific tool
```

qarmin (changes requested):
Except this two things, this is really good piece of code

czkawka_cli/src/commands.rs
MeirKlemp: I will add this:
```suggestion
    try "{usage} -h" to get more info about a specific tool
```

qarmin (approved):

qarmin:
Thanks!

czkawka_cli/src/main.rs
```diff
@@ -1,395 +1,182 @@
...
 
             #[cfg(not(debug_assertions))] // This will show too much probably unnecessary data to debug, comment line only if needed
             tf.print_results();
-
             tf.get_text_messages().print_messages();
         }
-        "--version" | "--v" => {
-            println!("Czkawka CLI {}", CZKAWKA_VERSION);
-            process::exit(0);
-        }
```
0x4A6F, line 353: Was the `--version` intentionally removed?
Is there another way to query for it?

MeirKlemp: It was probably left out by a mistake. It can be fixed by adding `version = CZKAWKA_VERSION` in here: https://github.com/qarmin/czkawka/blob/ae87489f3b52a4d1f9681f5ba71010a37550b9ef/czkawka_cli/src/commands.rs#L11
`CZKAWKA_VERSION` also needs to be imported.

PR #52, opened 2020-10-07
Improve code quality.

Caduser2020:
Fix clippy lints `-Wclippy::nursery`, decrease verbose constructs in the code, and generally attempt to decrease the complexity of functions and structs.

### DIFF
--- a/czkawka_core/src/big_file.rs
+++ b/czkawka_core/src/big_file.rs
@@ -151,7 +151,7 @@ impl BigFile {
                     }
                     for expression in &self.excluded_items.items {
                         if Common::regex_check(expression, &next_folder) {
-                            break 'dir;
+                            continue 'dir;
                         }
                     }
                     folders_to_check.push(next_folder);
@@ -184,7 +184,7 @@ impl BigFile {
                     // Checking expressions
                     for expression in &self.excluded_items.items {
                         if Common::regex_check(expression, &current_file_name) {
-                            break 'dir;
+                            continue 'dir;
                         }
                     }
 
--- a/czkawka_core/src/big_file.rs
+++ b/czkawka_core/src/big_file.rs
@@ -48,8 +48,8 @@ pub struct BigFile {
 }
 
 impl BigFile {
-    pub fn new() -> BigFile {
-        BigFile {
+    pub fn new() -> Self {
+        Self {
             text_messages: Default::default(),
             information: Info::new(),
             big_files: Default::default(),
@@ -67,15 +67,15 @@ impl BigFile {
         self.debug_print();
     }
 
-    pub fn get_big_files(&self) -> &BTreeMap<u64, Vec<FileEntry>> {
+    pub const fn get_big_files(&self) -> &BTreeMap<u64, Vec<FileEntry>> {
         &self.big_files
     }
 
-    pub fn get_text_messages(&self) -> &Messages {
+    pub const fn get_text_messages(&self) -> &Messages {
         &self.text_messages
     }
 
-    pub fn get_information(&self) -> &Info {
+    pub const fn get_information(&self) -> &Info {
         &self.information
     }
 
--- a/czkawka_core/src/big_file.rs
+++ b/czkawka_core/src/big_file.rs
@@ -163,12 +163,8 @@ impl BigFile {
                     .to_lowercase();
 
                     // Checking allowed extensions
-                    if !self.allowed_extensions.file_extensions.is_empty() {
-                        for extension in &self.allowed_extensions.file_extensions {
-                            if file_name_lowercase.ends_with((".".to_string() + extension.to_lowercase().as_str()).as_str()) {
-                                break;
-                            }
-                        }
+                    let allowed = self.allowed_extensions.file_extensions.iter().any(|e| file_name_lowercase.ends_with((".".to_string() + e.to_lowercase().as_str()).as_str()));
+                    if !allowed {
                         // Not an allowed extension, ignore it.
                         self.information.number_of_ignored_files += 1;
                         continue 'dir;
--- a/czkawka_core/src/big_file.rs
+++ b/czkawka_core/src/big_file.rs
@@ -151,7 +151,7 @@ impl BigFile {
                     }
                     for expression in &self.excluded_items.items {
                         if Common::regex_check(expression, &next_folder) {
-                            continue 'dir;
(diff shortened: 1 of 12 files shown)


Caduser2020:
Ready to review and test.

qarmin:
This PR needs to be rebased to current master, because I fixed(probably) issue with not testing always system builds like here.
Also clippy fails now.

But changes looks good, especially this one with default function(it was required by clippy, but it has almost zero usage)

czkawka_core/src/common.rs
```diff
@@ -7,9 +7,9 @@ use std::time::SystemTime;
 pub struct Common();
 impl Common {
     /// Printing time which took between start and stop point and prints also function name
-    pub fn print_time(_start_time: SystemTime, _end_time: SystemTime, _function_name: String) {
+    pub fn print_time(start_time: SystemTime, end_time: SystemTime, function_name: String) {
```
qarmin, line 11: I changed names to underscore because clippy or rust in release build talk that this variables are unused, but if this doesn't show now any warnings it will be a little strange.

ghost: Really, the function should be wrapped in a `#[cfg(debug_assertions)]`, but that would require wrapping all uses of it with `#[cfg(debug_assertions)]`. I would prefer doing that to having underscored variables used, but it is your decision.

qarmin: Putting in each line before `debug_print` a `#[cfg(debug_assertions)]`  is not intuitive, so I think that `#[allow(unused_variables)]` should be added to this function and underscores from variables should be also removed.

After this change CI should pass, and then I will able to check if there isn't any regressions.

ghost: Fixed in latest commit.

qarmin:
Finding temporary files is broken due not recognizing extensions properly
Zrzut ekranu z 2020-10-09 19-43-05

Caduser2020:
Should be fixed in latest commit.

qarmin:
But I write about temporary file finder.
Allowed extensions doesn't work fine(I didn't checked latest commit)

qarmin:
Also broken are excluded items when found any item to exclude(e.g. `*a*` can be used )

Caduser2020:
Very sorry, I think everything should be fixed in latest commit.

qarmin:
With excluded item `*g*` I have different result in master and this PR in duplicate finder
Zrzut ekranu z 2020-10-09 20-16-02

Caduser2020:
Failures are probably from commit 05b5310. Can you look over it to see what I messed up?

czkawka_core/src/big_file.rs
```diff
@@ -185,64 +164,53 @@ impl BigFile {
...
                             if file_name_lowercase.ends_with((".".to_string() + extension.to_lowercase().as_str()).as_str()) {
-                                have_valid_extension = true;
                                 break;
                             }
                         }
-                    } else {
-                        have_valid_extension = true;
+                        // Not an allowed extension, ignore it.
+                        self.information.number_of_ignored_files += 1;
+                        continue 'dir;
```
qarmin: Maybe I'm reading this bad, but if file have allowed extension, then always
```
self.information.number_of_ignored_files += 1;
                        continue 'dir;
```
is executed, but it shouldn't

qarmin:
I'm not sure but probably all `break 'dir` should be replaced with `continue 'dir`, because break stops checking other files/folders in dir

czkawka_core/src/duplicate.rs
```diff
@@ -292,16 +261,11 @@ impl DuplicateFinder {
                             };
 
                         // Checking expressions
-                        let mut found_expression: bool = false;
                         for expression in &self.excluded_items.items {
                             if Common::regex_check(expression, &current_file_name) {
-                                found_expression = true;
-                                break;
+                                break 'dir;
```
qarmin: This should be changed to `continue 'dir`
I know that `break` is in original code, but it is just a bug, because `'dir: for entry in read_dir ` iterates over all children and should be never break because it may skip some files/folders

czkawka_core/src/big_file.rs
qarmin: Also this seems to caused bug with temporary file finder

ghost: This is probably the bug. The break should break out of the if statement as well.

czkawka_core/src/duplicate.rs
ghost: I went ahead and reverted the attempted fixes which mistakenly added the `break` instead of the original `continue`.

Caduser2020:
Possibly fixed in latest commit.

czkawka_core/src/temporary.rs
```diff
@@ -176,88 +162,72 @@ impl Temporary {
...
                     if file_name_lowercase.contains('.') {
                         for temp in temporary_with_dot.iter() {
                             if file_name_lowercase.ends_with(temp) {
-                                is_temporary_file = true;
+                                break;
                             }
                         }
+                        self.information.number_of_ignored_files += 1;
+                        continue 'dir;
                     }
```
qarmin: When file doesn't contains dot, then should be ignored
This code works fine

```
                    let mut found : bool = false;

                    if file_name_lowercase.contains('.') {
                        for temp in temporary_with_dot.iter() {
                            if file_name_lowercase.ends_with(temp) {
                                 found = true;
                                 break;
                            }
                        }
                    }
                    if !found {
                        self.information.number_of_ignored_files += 1;
                        continue 'dir;
                    }

```

ghost: Rust prefers using `any()` over iterating and setting a flag. Would something like this work?
```rust
if !file_name_lowercase.contains('.') || !temporary_with_dot.iter().any(|f| file_name_lowercase.ends_with(f)) {
    self.information.number_of_ignored_files += 1;
    continue 'dir;
}

```

qarmin: Sure,
Tested it and works fine, and took only 4 lines instead 13 from my example.

ghost: Thanks, fixed in latest commit.

qarmin:
Thanks for this big refactoring!

Caduser2020:
Thanks! Happy to help.

PR #54, opened 2020-10-09
Update desktop file 

Tids:
Problem right now is that
1. If you launch the the GUI, the entry in your taskbar doesn't have an icon
2 if you have the desktop file pin'ed to you taskbar, the applaunch might create an additional icon

Both of them are a result, that the desktop can't know that the desktop file belongs to the app-window.
Both of them are fixed with `StartupWMClass=czkawka_gui`

3. if the binary is in a path that is *not* in $PATH, the applauncher will show up in the menu, but the click results in nothing.

This is fixed with `TryExec=czkawka_gui`

https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html


qarmin:
CC @sp1ritCS

sp1ritCS:
Wouldn't it be smarter to let GTK set the icon? [Gtk.Window.set_icon(string)](https://people.gnome.org/~shaunm/girdoc/C/Gtk.Window.set_icon.html)

Tids:
> Wouldn't it be smarter to let GTK set the icon?

The *.desktop file needs a hint what window belongs to the file, otherwise there would be 2 icons in some launcher-task manager combination bars. At least on wayland the desktop file icon is the used one after a window is bound to the desktop file.
So it might be a good idea to have it set in GTK as fallback and in the desktop file as main-source.

qarmin:
Thanks, this was merged in PR with updated icon.

Issue #44, opened 2020-10-06
Program crashes when processing files with (german) special chars


Lakrimar:
### Error Description
I wanted to search for duplicates but the program crashes when processing a file with a special char (ü).
I used czkawka version 1.0.0 

I also had FSLint installed and tried the same search with that to prevent hardware or filesystem issues and it worked fine. 

### Start Parameter
RUST_BACKTRACE=full ./czkawka_gui

### Stacktrace
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "\x81ber die Firma.txt"', czkawka_core/src/duplicate.rs:253:92
stack backtrace:
   0:     0x5573ba7259b5 - <unknown>
   1:     0x5573ba74979c - <unknown>
   2:     0x5573ba7239b2 - <unknown>
   3:     0x5573ba7282b0 - <unknown>
   4:     0x5573ba727ffc - <unknown>
   5:     0x5573ba7288f3 - <unknown>
   6:     0x5573ba7284eb - <unknown>
   7:     0x5573ba748131 - <unknown>
   8:     0x5573ba747f53 - <unknown>
   9:     0x5573ba6ed09c - <unknown>
  10:     0x5573ba6dda16 - <unknown>
  11:     0x7f698b113346 - <unknown>
  12:     0x7f698b12e9ff - g_signal_emit_valist
  13:     0x7f698b12f12f - g_signal_emit
  14:     0x7f698c4b0add - <unknown>
  15:     0x7f698c4b0b35 - <unknown>
  16:     0x7f698b113346 - <unknown>
  17:     0x7f698b12e9ff - g_signal_emit_valist
  18:     0x7f698b12f12f - g_signal_emit
  19:     0x7f698c4aef90 - <unknown>
  20:     0x7f69855f3dae - ffi_call_unix64
  21:     0x7f69855f371f - ffi_call
  22:     0x7f698b113ced - g_cclosure_marshal_generic_va
  23:     0x7f698b113346 - <unknown>
  24:     0x7f698b12e9ff - g_signal_emit_valist
  25:     0x7f698b12f12f - g_signal_emit
  26:     0x7f698c56ba36 - <unknown>
  27:     0x7f698b116008 - g_cclosure_marshal_VOID__BOXEDv
  28:     0x7f698b113346 - <unknown>
  29:     0x7f698b12e9ff - g_signal_emit_valist
  30:     0x7f698b12f12f - g_signal_emit
[...]

PR #59, opened 2020-10-10
Replace String with PathBuf for paths

darksv:
This PR changes handling of paths so that `String` is replaced with `PathBuf`. The latter is more suitable for this purpose as it is cross-platform and is not required to be valid UTF8. Also provides nice convenience methods like `extension` or `join` that correctly handles separators on various platforms.

### DIFF
--- a/czkawka_core/src/big_file.rs
+++ b/czkawka_core/src/big_file.rs
@@ -144,15 +144,10 @@ impl BigFile {
                     }
 
                     let next_folder = current_folder.join(entry_data.file_name());
-                    if self.directories.excluded_directories.contains(&next_folder) {
+                    if self.directories.is_excluded(&next_folder) || self.excluded_items.is_excluded(&next_folder) {
                         continue 'dir;
                     }
 
-                    for expression in &self.excluded_items.items {
-                        if Common::regex_check(expression, &next_folder) {
-                            continue 'dir;
-                        }
-                    }
                     folders_to_check.push(next_folder);
                 } else if metadata.is_file() {
                     // Extracting file extension
@@ -169,15 +164,9 @@ impl BigFile {
                     }
 
                     // Checking expressions
-                    let mut current_file_name = current_folder.join(entry_data.file_name());
-                    for expression in &self.excluded_items.items {
-                        if Common::regex_check(expression, &current_file_name) {
-                            continue 'dir;
-                        }
-                    }
-
-                    if cfg!(target_family = "windows") {
-                        current_file_name = Common::prettier_windows_path(&current_file_name);
+                    let current_file_name = current_folder.join(entry_data.file_name());
+                    if self.excluded_items.is_excluded(&current_file_name) {
+                        continue 'dir;
                     }
 
                     // Creating new file entry
--- a/czkawka_core/src/big_file.rs
+++ b/czkawka_core/src/big_file.rs
@@ -316,12 +316,14 @@ impl SaveResults for BigFile {
             }
         };
 
-        if let Err(_) = write!(
+        if writeln!(
             file,
-            "Results of searching {:?} with excluded directories {:?} and excluded items {:?}\n",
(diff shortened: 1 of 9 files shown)


qarmin:
Changes looks mostly good.

Seems that it even fixes not excluding folders like `.git` by default

Also probably this is better fix to #44

The only thing for now which is broken is saving to file, because it prints records with one empty line between them
```
---- Size 133.40 KiB (136606) - 112 files 
/home/rafal/Pobrane/home/runner/work/czkawka/czkawka/gtk_app/share/icons/Adwaita/cursors/dotbox.cur 

/home/rafal/Pobrane/home/runner/work/czkawka/czkawka/gtk_app/share/icons/Adwaita/cursors/cell.cur 

/home/rafal/Pobrane/home/runner/work/czkawka/czkawka/gtk_app/share/icons/Adwaita/cursors/row-resize.cur 

/home/rafal/Pobrane/home/runner/work/czkawka/czkawka/gtk_app/share/icons/Adwaita/cursors/ns-resize.cur 

/home/rafal/Pobrane/home/runner/work/czkawka/czkawka/gtk_app/share/icons/Adwaita/cursors/size_hor.cur 

/home/rafal/Pobrane/home/runner/work/czkawka/czkawka/gtk_app/share/icons/Adwaita/cursors/circle.cur 
```

qarmin:
Windows excluded directories are broken
Sample project - [Z.zip](https://github.com/qarmin/czkawka/files/5365456/Z.zip)

Executing command `.\czkawka_cli.exe dup -d "C:\Z" "c:/z" -e "c:/Z/Q"`
finds duplicates from folder `C:/z/q` which should be
```
Found 88 duplicated files in 1 groups with same content which took 12.77 MiB:
Size - 150.29 KiB (153901) - 88 files
C:/z/fffffffffffffffffffffffffffffffffffffffffffffffffffff
C:/z/fffffffffffffffffffffffffffffffffffffffffffffffffffff (10. kopia)
C:/z/q/fffffffffffffffffffffffffffffffffffffffffffffffffffff (31. kopia)
C:/z/q/fffffffffffffffffffffffffffffffffffffffffffffffffffff (32. kopia)
```

On Linux `./czkawka_cli dup -d /home/rafal/Pulpit/Z /home/rafal/Pulpit/Z /home/rafal/Pulpit/Z -e /home/rafal/Pulpit/Z/Q`
works as expected

czkawka_core/src/big_file.rs
```diff
@@ -144,59 +143,40 @@ impl BigFile {
...
                     for expression in &self.excluded_items.items {
                         if Common::regex_check(expression, &current_file_name) {
                             continue 'dir;
                         }
                     }
 
-                    #[cfg(target_family = "windows")]
-                    {
+                    if cfg!(target_family = "windows") {
                         current_file_name = Common::prettier_windows_path(&current_file_name);
```
qarmin: `prettier_windows_path` function was used only for normalizing path like `c:/Av`, `c:\AV`, `C:/av` to `C:/av` to allow easily remove duplicate path or overlap ones.
With PathBuf I don't think that this is necessary, and can be removed(of course if everything else will work fine)

qarmin:
Pathbuf doesn't recognize that on Windows
```
"c:/z\\Q" is equal to :"c:/z/q"
``` 
and the only solution for me is to check every single time when using excluding directories if string path converted to lowerspace are equal

darksv:
Right, I'll fix this when I got some time.

qarmin:
You can try to rebase to current master, because I added recently CI job with checking simple regressions and it probably will allow you to test PR without needing using Windows or Wine(The longest job takes 6/7 minutes).

darksv:
I have extracted some duplicated code related to exclusion into a separated methods. Please check whether it's OK.

qarmin (approved):

qarmin:
Seems that everything works now.
Thanks, especially for simplifying code!

PR #72, opened 2020-10-15
Make image similarity search significantly faster

Syfaro:
This PR makes the image similarity search significantly faster. 

It does this by comparing the hamming distance of 8-byte arrays instead of the Levenshtein distance of base64 encoded data.

I ran master and this revision (both in release mode) against 10,312 files in 9,619 folders. The current master took around 410 seconds to complete, this version takes around 70 seconds.


qarmin (approved):

qarmin:
Thanks!
With my test collection time dropped from 121 seconds to 55 seconds.

PR #122, opened 2020-12-19
Add button to enable/disable in settings text view errors

qarmin:


codywalker77 (commented):
All codes wettest correct and true

Issue #23, opened 2020-10-02
Add cache for duplicate finder


qarmin:
For now if user twice check duplicates, then hash of files which still have same size needs to be checked again which is a little slow. 
This option should be switchable from GUI.

qarmin:
Should be added also an option to turn on/off this config

PR #139, opened 2020-12-27
Add cache support for similar images

qarmin:
Implements #23 for similar images

### DIFF
--- a/czkawka_core/src/similar_images.rs
+++ b/czkawka_core/src/similar_images.rs
@@ -5,19 +5,27 @@ use crate::common_messages::Messages;
 use crate::common_traits::{DebugPrint, PrintResults, SaveResults};
 use bk_tree::BKTree;
 use crossbeam_channel::Receiver;
+use directories_next::ProjectDirs;
 use humansize::{file_size_opts as options, FileSize};
 use image::GenericImageView;
 use img_hash::HasherConfig;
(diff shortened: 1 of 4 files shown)


czkawka_core/src/similar_images.rs
```diff
@@ -613,3 +667,103 @@ fn get_string_from_similarity(similarity: &Similarity) -> &str {
         Similarity::None => panic!(),
     }
 }
+
+fn save_hashes_to_file(hashmap: &HashMap<String, FileEntry>, text_messages: &mut Messages) {
+    if let Some(proj_dirs) = ProjectDirs::from("pl", "Qarmin", "Czkawka") {
+        // Lin: /home/alice/.config/barapp
```
alexisLefebvre, line 673: What is `barapp`? The `~/.config/barapp` doesn't exist, even if it looks like the cache works.

alexisLefebvre: Only this comment looks wrong, the rest of the code and the cache work! :+1:

qarmin: This comment was intended to show only where this function want to save data in general(comment was copied from `directories-next` crate).
Proper locations I put to instruction.
I have updated this comments in https://github.com/qarmin/czkawka/issues/196

Issue #196, opened 2021-01-11
Linux: Store cache in the `~/.cache/czkawka` directory


alexislefebvre:
Since #139 (which was a great addition!), the hashes of similar images are stored in a cache.

On Linux, the cache is stored in `~/.config/czkawka`: https://github.com/qarmin/czkawka/blob/9a0432d237b21f6a7951bc6d6a9f06c5a3047bd1/instructions/Instruction.md#configcache-files

I think it would be better to store the cache in `~/.cache/czkawka`. The freedesktop spec recommend to use it:

> `$XDG_CACHE_HOME` defines the base directory relative to which user specific non-essential data files should be stored. If `$XDG_CACHE_HOME` is either not set or empty, a default equal to `$HOME/.cache` should be used. 

Source: https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html

PR #197, opened 2021-01-11
Move image cache to cache from config dir

qarmin:
Fixes https://github.com/qarmin/czkawka/issues/196

I was think before about putting image cache to cache dir, but I want to hav everything at one place.
Probably it is better to put it to special localization.

Only Czkawka 2.2 will have an option to move image cache from config dirs to cache dirs.


alexislefebvre:
That's nice. Thanks!

PR #234, opened 2021-01-30
Ignore duplicates if those are hard links

blob79:
This is a proof of concept.

```
$ echo a > hardlinks/a
$ cp hardlinks/{a,b}
$ ln hardlinks/{a,c}
$ cargo run --bin czkawka_cli dup -m 1 --directories $(pwd)/hardlinks -f /dev/stderr > /dev/null
-------------------------------------------------Files with same hashes-------------------------------------------------
Found 1 duplicated files which in 1 groups which takes 2 B.

---- Size 2 B (2) - 2 files
/home/thomas/Development/czkawka/hardlinks/a
/home/thomas/Development/czkawka/hardlinks/b
```

Open:
- Windows support
- Probably this should be a cli option

### DIFF
--- a/czkawka_core/src/duplicate.rs
+++ b/czkawka_core/src/duplicate.rs
@@ -1,9 +1,13 @@
 use crossbeam_channel::Receiver;
 use humansize::{file_size_opts as options, FileSize};
+#[cfg(target_family = "unix")]
+use std::collections::HashSet;
 use std::collections::{BTreeMap, HashMap};
 use std::fs::{File, Metadata, OpenOptions};
 use std::io::prelude::*;
 use std::io::{Error, ErrorKind, Result};
+#[cfg(target_family = "unix")]
+use std::os::unix::fs::MetadataExt;
 use std::path::{Path, PathBuf};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use std::{fs, thread};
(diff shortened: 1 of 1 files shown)


czkawka_core/src/duplicate.rs
```diff
@@ -905,9 +909,10 @@ impl DuplicateFinder {
             self.information.bytes_read_when_hashing += bytes_read;
             self.text_messages.warnings.append(&mut errors);
             for (_hash, vec_file_entry) in hash_map {
-                if vec_file_entry.len() > 1 {
+                let dups = self.filter_hard_links(vec_file_entry);
+                if dups.len() > 1 {
                     self.files_with_identical_hashes.entry(size).or_insert_with(Vec::new);
-                    self.files_with_identical_hashes.get_mut(&size).unwrap().push(vec_file_entry);
+                    self.files_with_identical_hashes.get_mut(&size).unwrap().push(dups);
```
qarmin, line 915: This should be moved into `check_file_size` function, because for now this only works for full file hashing and checking files before hashing will allow to drop some records before doing such resource heavy operation.

qarmin (approved):

czkawka_core/src/duplicate.rs
qarmin: This should be moved into `check_file_size` function, because for now this only works for full file hashing and checking files before hashing will allow to drop some records before doing such resource heavy operation.

qarmin (changes requested):

Sbgodin:
Thanks for this improvement, I guess it would be also nice to choose whether it will be ignored or not.

qarmin:
Thanks!

PR #236, opened 2021-01-31
Add delete method to replace duplicate files with hard links

blob79:
This results in the same space saving but keeps the "deleted" files around.

$ mkdir hardlink
$ cd hardlink
$ echo a > a
$ cp a b
$ ln a c
$ touch -t 01010000 a
$ ls -il --time-style=full-iso .
25169323 -rw-rw-r-- 2 thomas thomas 2 2021-01-01 00:00:00.000000000 +0100 a
25169558 -rw-rw-r-- 1 thomas thomas 2 2021-01-31 18:11:08.298161098 +0100 b
25169323 -rw-rw-r-- 2 thomas thomas 2 2021-01-01 00:00:00.000000000 +0100 c

$ czkawka_cli dup --directories $(pwd)/hardlink -m 1 -f hardlink.result --delete-method HARD

$ ls --time-style=full-iso -li hardlink
25169323 -rw-rw-r-- 3 thomas thomas 2 2021-01-01 00:00:00.000000000 +0100 a
25169323 -rw-rw-r-- 3 thomas thomas 2 2021-01-01 00:00:00.000000000 +0100 b
25169323 -rw-rw-r-- 3 thomas thomas 2 2021-01-01 00:00:00.000000000 +0100 c


qarmin (approved):

qarmin:
Thanks!

PR #253, opened 2021-02-07
Keep original file if replacing duplicate with hardlink fails

blob79:
If the user could not create a hard link (due to permissions or
different block devices) then czkawka lost the duplicate file.

```
$ mkdir hardlink
$ cd hardlink
$ echo a > a
$ cp a b
$ chown thetestuser:thetestuser a
$ ls -il .

25169246 -rw-rw-r-- 1 thetestuser thetestuser 2 Feb  7 11:18 a
25169641 -rw-rw-r-- 1 thomas      thomas      2 Feb  7 11:18 b

$ cargo run --bin czkawka_cli dup --directories $(pwd) -m 1 -f
test101.result --delete-method HARD

[...]
-------------------------------WARNINGS--------------------------------
Failed to link /home/thomas/Development/czkawka/hardlink/b ->
/home/thomas/Development/czkawka/hardlink/a
---------------------------END OF WARNINGS-----------------------------

$ ls -il .
25169246 -rw-rw-r-- 1 thetestuser thetestuser 2 Feb  7 11:18 a
```

Now czkawka keeps all files and the warning provides more information
why czkawka can't replace the duplicate with a hard link.

```
$ cargo run --bin czkawka_cli dup --directories $(pwd) -m 1 -f
test101.result --delete-method HARD

-------------------------------WARNINGS--------------------------------
Failed to link /home/thomas/Development/czkawka/hardlink/b ->
/home/thomas/Development/czkawka/hardlink/a (Operation not permitted (os
error 1))
---------------------------END OF WARNINGS-----------------------------
[...]

25169246 -rw-rw-r-- 1 thetestuser thetestuser 2 Feb  7 11:18 a
25169641 -rw-rw-r-- 1 thomas      thomas      2 Feb  7 11:18 b
```

### DIFF
--- a/czkawka_core/src/duplicate.rs
+++ b/czkawka_core/src/duplicate.rs
@@ -3,6 +3,7 @@ use humansize::{file_size_opts as options, FileSize};
 use std::collections::{BTreeMap, HashMap};
 use std::fs::{File, Metadata, OpenOptions};
 use std::io::prelude::*;
+use std::io::{Error, ErrorKind, Result};
 use std::path::{Path, PathBuf};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use std::{fs, thread};
@@ -1280,12 +1281,12 @@ fn delete_files(vector: &[FileEntry], delete_method: &DeleteMethod, warnings: &m
             let src = vector[q_index].path.clone();
             for (index, file) in vector.iter().enumerate() {
                 if q_index != index {
-                    if fs::remove_file(file.path.clone()).and_then(|_| fs::hard_link(&src, &file.path)).is_ok() {
+                    if let Err(e) = make_hard_link(&src, &file.path) {
+                        failed_to_remove_files += 1;
+                        warnings.push(format!("Failed to link {} -> {} ({})", file.path.display(), src.display(), e));
+                    } else {
                         removed_files += 1;
                         gained_space += file.size;
-                    } else {
-                        failed_to_remove_files += 1;
-                        warnings.push(format!("Failed to link {} -> {}", file.path.display(), src.display()));
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


qarmin (approved):

qarmin:
Thanks!

czkawka_core/src/duplicate.rs
```diff
@@ -1297,6 +1298,17 @@ fn delete_files(vector: &[FileEntry], delete_method: &DeleteMethod, warnings: &m
     (gained_space, removed_files, failed_to_remove_files)
 }
 
+fn make_hard_link(src: &PathBuf, dst: &PathBuf) -> Result<()> {
+    let dst_dir = dst.parent().ok_or_else(|| Error::new(ErrorKind::Other, "No parent"))?;
+    let temp = tempfile::Builder::new().tempfile_in(dst_dir)?;
+    fs::rename(dst, temp.path())?;
```
krzysdz, line 1304: On Windows it's not possible to rename a file if the "target" file exists. Because the `tempfile_in` function creates a file (not just a name/path), the rename will fail with a permission denied error.

```
Error: Os { code: 5, kind: PermissionDenied, message: "Odmowa dostępu." }
```

czkawka_core/src/duplicate.rs
```diff
@@ -1600,3 +1612,61 @@ fn load_hashes_from_file(text_messages: &mut Messages, type_of_hash: &HashType)
...
     None
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs::{read_dir, File};
+    use std::io::Result;
+    #[cfg(target_family = "windows")]
+    use std::os::fs::MetadataExt;
```
krzysdz, line 1622: It should be `use std::os::windows::fs::MetadataExt;`, but the import (only on Windows) is unused - the `assert_inode` function is empty on Windows and it's the only place, where `MetadataExt` is used (inode number check).

czkawka_core/src/duplicate.rs
blob79: I've no windows machine. The documentation says https://doc.rust-lang.org/std/fs/fn.rename.html:
> This function currently corresponds to the rename function on Unix and the MoveFileEx function with the MOVEFILE_REPLACE_EXISTING flag on Windows.

I assume it's actually a different root cause, e.g. that the files are on different file systems. You could try to do the same file operations manually and see where it fails.

czkawka_core/src/duplicate.rs
blob79: I suppose you could do a pull request.

Issue #176, opened 2020-12-31
Show scan progression in Windows bar.


goodboi69:
Show the progression in the Windows bar like other programs.
Example: Downloads from Firefox/Chrome/etc.

![Untitled-1](https://user-images.githubusercontent.com/43517545/103426395-33ffc380-4b87-11eb-9acf-00b5dcea11f6.jpg)

qarmin:
Fixed by #264

PR #264, opened 2021-02-14
Windows taskbar progress support

krzysdz:
This PR adds support for taskbar progress (#176) on Windows (7+). It should be also possible to implement it for some Linux systems using [libxapp](https://github.com/linuxmint/xapp), but I did not find Rust bindings for it.

However, there are some problems to solve and things to improve:

- [ ] Taskbar progress is not available before Windows 7 and I have no idea what will happen on earlier versions. Runtime version checks could be added (using [VerifyVersionInfoA](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-verifyversioninfoa)) if the GUI is supposed to run on older Windows versions.
- [x] If error and pause states will not be used (not supported in XApp anyway), setting the normal progress state is not necessary (setting value changes the state to normal from indeterminate and no progress).
- [x] **IMPORTANT:** The struct is never dropped and COM is not uninitialised properly. It looks like the clones of `GuiData` and `Rc<TaskbarProgress>` moved to closures used as signal handlers or spawned in the main context [are not dropped](https://gist.github.com/krzysdz/6f28f7938b33a0e7210b7a09dcd7a4f9#file-not_dropped-txt).

Comments and suggestions are welcome.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -600,6 +600,7 @@ dependencies = [
  "humansize",
  "image",
  "open",
+ "winapi",
 ]
 
 [[package]]
--- a/czkawka_gui/Cargo.toml
+++ b/czkawka_gui/Cargo.toml
@@ -32,7 +32,7 @@ open = "1.4.0"
 image = "0.23.12"
 
 [target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3.9", features = ["combaseapi", "guiddef", "objbase", "shobjidl_core", "unknwnbase", "windef", "winerror", "wtypesbase", "winuser"] }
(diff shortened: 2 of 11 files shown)


alexislefebvre:
> Taskbar progress is not available before Windows 7 and I have no idea what will happen on earlier versions

Windows XP and Vista are not supported anymore, I think it's acceptable to not support these outdated versions.

krzysdz:
I have also some tests in krzysdz/czkawka#1, which I can add if you want, but I think this PR is now ready for review.

qarmin:
Thanks!
Changes looks good, progress bar works, I don't see any crashes etc.

PR #265, opened 2021-02-17
Fix Winows theme in CI builds

krzysdz:
The `admin-gtk3-dark-osx` theme contains relative symlinks, which after copying only this theme are no longer valid.\
The **-L** (**--dereference**) option makes `cp` always follow symbolic links, so that all of them are replaced with files they pointed to.

The `zip` command ignored these invalid symlinks (`zip warning: name not matched` in [logs](https://github.com/qarmin/czkawka/pull/264/checks?check_run_id=1915425586#step:7:44482)) and the CI builds didn't contain the `assets/`, `windows-assets/` directories and `assets.svg`, `assets.txt` files.

This fixes nrhodes91/AdMin#11.


krzysdz:
Before (3ad1dfa, artifacts from [this run](https://github.com/qarmin/czkawka/runs/1881576818)) and after (988ce92, artifacts from [this run](https://github.com/qarmin/czkawka/runs/1922569313)) comparison.

Before and after - checkboxes and window icons are now fixed

qarmin (approved):

qarmin:
Thanks!

PR #268, opened 2021-02-21
Use one implementation for all hash calculations

blob79:
All implementations are the same they only differ by hasher and the file read
limit.


qarmin:
Tested and looks that works fine,
Thanks!

Issue #187, opened 2021-01-09
Pressing the “Delete” button doesn't trigger a warning


alexislefebvre:
Thanks for this great software!

I'm using the “Similar Images” with the 2.1.0 version, on Linux

If I use the “Delete” button in the app, there is a dialog:

> ![image](https://user-images.githubusercontent.com/2071331/104110280-8606be80-52d6-11eb-85b0-9f1aa86c1dc1.png)

But If I use the `Delete` key of my keyboard, there is no dialog.

Could we have the same warning in both cases?

dandv:
I've also run into this, and was surprised to see nothing happen after I pressed the <kbd>Delete</kbd> key.

If it's intentional that pressing <kbd>Delete</kbd> should not prompt for confirmation, then it could display a confirmation message, like "File xxx has been deleted".

PR #282, opened 2021-03-06
Add confirmation dialog when removing files with delete key

### LINKED ISSUE #259: Suggestion: set the focus on the OK button in the Delete dialog
As a safe but still fast workflow, I click Delete on files I want to delete, then I could press Enter. Since I'm right handed, the mouse is near the Enter key on the numpad, so this is fast.

The Problem is that pressing Enter toggles the "Ask in future" checkbox:

image

### LINKED ISSUE #279: <space> on progress window cancels progress
Waking up my computer with a space key has killed many hour runs twice now.  Is there a mapping for the the 'Stop' button that would be less sensitive?

qarmin:
Fixes https://github.com/qarmin/czkawka/issues/187
Fixes #259
Fixes #279

I used workaround for GTK problem - https://github.com/gtk-rs/gtk-rs/issues/336

### DIFF
--- a/czkawka_gui/czkawka.glade
+++ b/czkawka_gui/czkawka.glade
@@ -2586,7 +2586,7 @@ This program is free to use and will always be.
                 <property name="orientation">vertical</property>
                 <child>
                   <object class="GtkCheckButton" id="check_button_settings_hide_hard_links">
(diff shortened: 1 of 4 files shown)


alexislefebvre:
Thanks for working on this!

PR #283, opened 2021-03-08
Use xxhash-rust >=0.8.2 to avoid bug in stateful implementation

DoumanAsh:
Unfortunately stateful `Xxh3` had bug which resulted in it producing invalid hashes for total length of 1025 and more.
0.8.1 is not a valid version to use now


qarmin:
Thanks!

Issue #133, opened 2020-12-23
Add multithread support for collecting files for search


qarmin:
Most things support multithreading but this no.

PR #314, opened 2021-04-11
[Very early WIP] Use multithreading to collect files to scan

qarmin:
Attempt to implement https://github.com/qarmin/czkawka/issues/133, looks that it may be very hard to achieve performance improvement

### DIFF
--- a/czkawka_core/src/duplicate.rs
+++ b/czkawka_core/src/duplicate.rs
@@ -18,6 +18,7 @@ use std::{fs, mem, thread};
 use crossbeam_channel::Receiver;
 use directories_next::ProjectDirs;
 use humansize::{file_size_opts as options, FileSize};
+use ignore::WalkBuilder;
 use rayon::prelude::*;
 
 use crate::common::Common;
@@ -185,10 +186,11 @@ impl DuplicateFinder {
 
         match self.check_method {
             CheckingMethod::Name => {
-                if !self.check_files_name(stop_receiver, progress_sender) {
-                    self.stopped_search = true;
-                    return;
+                for _ in 0..5 {
+                    self.check_files_name(stop_receiver, progress_sender);
+                    self.check_files_name_threaded(stop_receiver, progress_sender);
                 }
+                std::process::exit(0);
             }
             CheckingMethod::Size => {
                 if !self.check_files_size(stop_receiver, progress_sender) {
@@ -314,7 +316,7 @@ impl DuplicateFinder {
         self.excluded_items.set_excluded_items(excluded_items, &mut self.text_messages);
     }
 
-    fn check_files_name(&mut self, stop_receiver: Option<&Receiver<()>>, progress_sender: Option<&futures::channel::mpsc::UnboundedSender<ProgressData>>) -> bool {
+    fn check_files_name(&mut self, stop_receiver: Option<&Receiver<()>>, progress_sender: Option<&futures::channel::mpsc::UnboundedSender<ProgressData>>) {
         let start_time: SystemTime = SystemTime::now();
         let mut folders_to_check: Vec<PathBuf> = Vec::with_capacity(1024 * 2); // This should be small enough too not see to big difference and big enough to store most of paths without needing to resize vector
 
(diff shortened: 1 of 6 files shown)


alexislefebvre:
Is the filesystem the bottleneck?

qarmin:
For now when collecting files to scan on SSD and even HDD, CPU is bottleneck because in same thread it read file metadata(which depends on disk I/0) and later execute several functions this metadata to check if file/folder is proper.

According to https://pkolaczk.github.io/disk-parallelism/, SSD and HDD will gain big performance boost with random data access(which happens when reading files metadata not content)

MrJake222:
Also calculating partial/full hashes could be multithreaded I think. CPU is a bottleneck there also.

qarmin:
> Also calculating partial/full hashes could be multithreaded I think. CPU is a bottleneck there also.

This was implemented in https://github.com/qarmin/czkawka/pull/100

czkawka_core/src/duplicate.rs
```diff
@@ -344,112 +340,156 @@ impl DuplicateFinder {
...
-                    if !self.recursive_search {
-                        continue 'dir;
-                    }
 
-                    let next_folder = current_folder.join(entry_data.file_name());
-                    if self.directories.is_excluded(&next_folder) {
-                        continue 'dir;
+                // Checking allowed extensions
+                if !allowed_extensions.is_empty() {
+                    let allowed = allowed_extensions.iter().any(|e| file_name_lowercase.ends_with((".".to_string() + e.to_lowercase().as_str()).as_str()));
```
pczarn, line 552: I believe this line has too much complexity. You may lift it by moving code to the `Extensions` abstraction like this:

```rust
impl Extensions {
    fn contains(&self, file_name_lowercase: &str) -> bool {
        self.file_extensions.iter().any(|e| {
            let dot_then_ext = ".".to_string() + e.to_lowercase().as_str();
            file_name_lowercase.ends_with(dot_then_ext.as_str())
        })
    }
}

// ...
if !self.allowed_extensions.contains(&file_name_lowercase[..]) {
    // Not an allowed extension, ignore it.
    continue;
}
```

pczarn (approved):
Looks good to me except for that one comment.

czkawka_core/src/duplicate.rs
pczarn: my bad, use a more descriptive function name like `matches_filename`, not `contains`

qarmin: This could be optimized, by adding when necessary dots at the start and use `to_lowercase` in `set_allowed_extensions` function.

But this should be done in new PR, because I'm not sure when I merge this PR, because in 2 implementations of threads which I wrote, I got smaller performance than without.

pczarn: I could volunteer to do such a simple task.

pczarn: I feel like this is premature optimization -- on the other hand, this code is O(m*n) in the number of files (m) times the number of allowed extensions (n) so we could use some kind of set data structure here.

pczarn: So this is a low hanging fruit up for optimization.

pczarn:
@qarmin Send the code my friend, and I will look into performance. How do you benchmark this thing?

//

Podeślij kod, kumplu, to zobaczymy jak z przetwarzaniem równoległym. Jak to benchmarkujesz? Nie widzę w repo żadnych skryptów do benchmarkowania, może warto je napisać? Mam tylko SDD, bez HDD, ale może coś zaradzę.

qarmin:
To benchmark this, I used 2 apps compiled with different implementations of algorithm, but this was quite invalid approach so now after rebase it runs 2 algorithm in one app run.

To benchmark app, it is required to execute this command
```
cargo run --bin czkawka_cli --release  -- dup -s NAME -d /mnt/Miecz/
```
where `/mnt/Miecz` is directory to check(in my example 108000 files which takes 185 GB on SSD).
Output like this should show:
```
SEARCHING NORMAL TOOK 345.273014ms, found 27334 files
SEARCHING THREADED TOOK 688.543524ms, found 40856 files
SEARCHING NORMAL TOOK 337.022304ms, found 27334 files
SEARCHING THREADED TOOK 696.532427ms, found 40856 files
```
Number of found files should be similar, but probably I missed some checks.

PR #327, opened 2021-05-09
Update Installation.md

tenninjas:
The Mac instructions need to be clear that you are telling people to install Homebrew first which will manage GTK; People should be informed about what you are asking them to install on their system. Here is some example wording that I think helps with that - though I didn't spend more than a few seconds on it and you may want to word it differently.


alexislefebvre (approved):

qarmin:
Thanks!

Issue #338, opened 2021-05-13
GTK icons and theme problems in appimage


ShrirajHegde:
Misbehaving icons and theme [appimage]

![image](https://user-images.githubusercontent.com/66936172/118097163-f2faa400-b3ef-11eb-9d09-3ed7f1a0747a.png)

In standalone binary:

![image](https://user-images.githubusercontent.com/66936172/118097243-1291cc80-b3f0-11eb-908e-5c148ca5b279.png)

qarmin:
This is not a bug, gtk plugin for appimage ships own theme

PR #339, opened 2021-05-13
Generate Minimal Appimage

ShrirajHegde:
Generate Minimal Appimage with stripped binary and no GTK  included. Reduces size to under 2 MB

solves #338


qarmin:
Thanks!

PR #160, opened 2020-12-29
Adds Mac GUI

qarmin:
Fixes https://github.com/qarmin/czkawka/issues/11
Fixes https://github.com/qarmin/czkawka/issues/64

Thanks to ImplicitEmpiricism I have now basic instructions how to compile Czkawka on MacOS.
```
xcode-select --install
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
brew install rust gtk+3 
curl -L https://github.com/qarmin/czkawka/archive/2.0.0.tar.gz | tar -zxvf -
cd czkawka-2.0.0
cargo run --bin czkawka_gui
```
Let's see if I can wrap it inside Github CI

Test binary - [czkawka_gui-macOS-stable.zip](https://github.com/qarmin/czkawka/files/5751893/czkawka_gui-macOS-stable.zip)

It should be able to run with help of terminal.
Probably after some changes it will be able to run when double clicking at it.

If you could open app, then please print here screenshot with OS info,

PR #368, opened 2021-05-23
Add adwaita-icon-theme as required dependency on macOS

NickKaramoff:
Apparently, it's not enough to install the `gtk+3` package on macOS — the icons won't show up. One needs to install the `adwaita-icon-theme` as a separate package. Using the app without icons is possible, but not the greatest experience.

----

If one doesn't install it, the following message will appear in Terminal:

```
(<unknown>:****): Gtk-WARNING **: 12:34:56.789: Could not load a pixbuf from /org/gtk/libgtk/theme/Adwaita/assets/check-symbolic.svg.
This may indicate that pixbuf loaders or the mime database could not be found.
```

### DIFF
--- a/instructions/Installation.md
+++ b/instructions/Installation.md
@@ -19,11 +19,11 @@ sudo xbps-install gcc pkg-config alsa-lib-devel
 ```
 
 ### macOS
-Currently, you need to manually install `GTK 3` libraries, because they are dynamically loaded from the OS (*we need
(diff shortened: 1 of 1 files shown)


qarmin:
It works?
When I added support for mac in https://github.com/qarmin/czkawka/pull/160#issuecomment-753885589, then probably this solution doesn't work.

NickKaramoff:
Huh, that's interesting. It did work for me 🤔🤷‍♂️

qarmin:
Thanks!

PR #396, opened 2021-07-22
Add webp support to similar images

tecome:


qarmin:
This should be added also here 
https://github.com/qarmin/czkawka/blob/4fc86552726c109878d7327782bf285bb1f73c6b/czkawka_core/src/broken_files.rs#L682

I remember that last time when I tested it, then it shows a big number about non supported colours/type

tecome:
`.webp` is already in `allowed_image_extensions` unless I'm missing something. I've tested it and it's working properly

qarmin:
Yeah, I missed that(I'm probably blind)

qarmin:
Thanks

PR #415, opened 2021-08-11
Use space key to select files in search result

kuskov:
It's more convenient than using mouse when we're dealing with a lot of files.


qarmin:
Thanks!
I wanted to do this before releasing 3.2 version, but I forgot to do this.

For now I see that clicking space remove selection entire selection except the latest selected row, but it happens also with 3.1.0 version so this must be proper gtk behavior(later this could be disabled)

PR #427, opened 2021-08-20
Cargo update (update dependencies)

sigaloid:
(does pass tests btw)


qarmin:
Thanks, I already updated dependencies manually in other commit because I needed to make sure that new requirements works with newly added to CI Rust 1.53.0

PR #433, opened 2021-09-22
remove default excluded directories and items

thomas725:
by doing so, when having the two lines:
```
--load_at_start:
false
```
in the config file @ `~/.config/czkawka/czkawka_gui_config.txt` we can make `czkawka_gui` use the working directory as the initial included directory.


qarmin:
Most of users expect that by default such directories will be excluded.
This probably should be done by simply adding `use_empty_included_directories` or `dont_load_excluded_directories`

qarmin:
Implemented by  #677

PR #461, opened 2021-11-23
Remove support for zeroed files

qarmin:
Looks that not zeroed files were not used too often, so it is better to remove this and focus on other functions.
This will also give more place in GUI, because after adding SimilarVideos, there was at left side a little too much things
I think that most of cases of zeroed files could be catched by broken files tool


themespan:
I use this feature frequently. Please consider putting it back.

alexislefebvre:
This can be done from CLI: `find . -type f -empty -delete` ([source](https://stackoverflow.com/questions/5475905/linux-delete-file-with-size-0/18052644#18052644))

themespan:
I'm on Windows, I don't think that command finds zeroed files. Czkawka has a nice GUI around this functionality and the code is already written, it would be a shame to remove this very useful feature.

PR #486, opened 2021-12-04
Cache vidoes with serde to avoid forked vid_dup_finder_lib

Farmadupe:
[Issue #1 in vid_dup_finder_lib](https://github.com/Farmadupe/vid_dup_finder_lib/issues/1) requests to add additional data to the API of vid_dup_finder_lib to serialize VideoHash, however it is already possible to serialize VideoHash using Serde.

Main Changes:
- similar_videos.rs is updated to serialize data to JSON instead of custom format.
- cargo.toml is updated to use vid_dup_finder_lib from crates.io

Other changes:
- In existing functionality, if there was an error on any single line of the cache file, then only that line will fail. In this change, the whole serialization will fail. However I think it is unlikely to experience such an error (unless the user manually edits this file)

Thoughts:
- I chose to use JSON serialization format because the existing format is human-readable and JSON is too. But with large cache it will be slow to serialize. Maybe a binary format would be better (such as bincode)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -493,6 +493,8 @@ dependencies = [
  "img_hash",
  "rayon",
  "rodio",
+ "serde",
+ "serde_json",
  "tempfile",
  "vid_dup_finder_lib",
  "xxhash-rust",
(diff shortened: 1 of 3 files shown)


qarmin:
Looking at https://davidkoloski.me/blog/rkyv-is-faster-than/ I see that serde_json can be a lot of slower than bincode, so probably usage of bincode would be better idea(at least for release build, for debug serde_json can help with debugging problems)

Also 140KB of cache with 180 results is a little too much for me, because users can store even hundread of thousands records so every second/byte matters.(Old implementation takes 70KB for 180 files, but not sure how about loading times)

I miss my old custom implementation, but looks that this is proper way to do cache system(a lot of other things also is not written in optimal/proper way in this app, e.g. GTK developers advised me to rewrite the application, because I shouldn't create a GUI in that way XD)

Farmadupe:
Compared to the terrible GUI I made for vid_dup_finder, the GUI for czkawka is perfect, I am not complaining!

In the latest commit, JSON is replaced with bincode. I did not perform a benchmark with a large dataset, but I am sure that it would be much faster.

qarmin:
Thanks

PR #502, opened 2021-12-15
Parallel BFS with Rayon

pczarn:
Found almost 3x speedup with Rayon (my computer has 4 cores)

Code based on code from stackexchange post https://codereview.stackexchange.com/questions/270874/parallel-breadth-first-traversal-in-rust

### DIFF
--- a/czkawka_core/src/duplicate.rs
+++ b/czkawka_core/src/duplicate.rs
@@ -469,9 +469,9 @@ impl DuplicateFinder {
             folders_to_check.clear();
 
             // Process collected data
-            for (mut segment, mut warnings, fe_result) in segments {
-                folders_to_check.append(&mut segment);
-                self.text_messages.warnings.append(&mut warnings);
+            for (segment, warnings, fe_result) in segments {
+                folders_to_check.extend(segment);
+                self.text_messages.warnings.extend(warnings);
                 for (name, fe) in fe_result {
                     self.files_with_identical_names.entry(name.clone()).or_insert_with(Vec::new);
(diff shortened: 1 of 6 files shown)


czkawka_core/src/duplicate.rs
```diff
@@ -355,109 +356,136 @@ impl DuplicateFinder {
 
         //// PROGRESS THREAD END
 
-        while !folders_to_check.is_empty() {
+        let mut bounds: Range<usize> = 0..folders_to_check.len();
```
qarmin: Why range is needed here?

Instead `bounds` and `current_slice` I used directly `folders_to_check` and it compiles fine
Only before `// Process collected data` I need to put `folders_to_check = vec![]`

pczarn: @qarmin I see: it's not needed. The new commit does `folders_to_check.clear()`. I coped a few lines of code and didn't realize it can be simpler.

pczarn:
We have 2 new commits

qarmin:
Dzięki!

Looks that with i7-4770(4/8 HT) I got usually 2,5-3,5x speedup on cheap SSD and ~2x speedup on HDD.
Implementation is a lot of simpler that I expected and is really easy to understand.

alexislefebvre:
Could you please explain what will be faster? Indexing? Comparing images?

qarmin:
This is improved version of https://github.com/qarmin/czkawka/pull/314, so faster will be searching for files.
For now it is implemented only in one mode(duplicates by name), but I will add support for other modes soon.

Issue #232, opened 2021-01-28
Add notion of reference directory


nirvdrum:
I'm trying to clear out space on an SD card that had photos and videos on it. Over the years I've imported data from it, but the importing tool didn't clear things off after the import. So, I've ended up with duplicates on subsequent imports. It's great that those are found, too, since I ultimately would like to get rid of all the duplicates, but it makes it hard to clear off the SD card when the results are mixed in with duplicates found for file not on the SD card.

It would be helpful if I could annotate one directory as a reference or source directory and have duplicates only reported if they're also present in the filesystem tree rooted at the referenced directory.

ca-boh:
Yep. Czkawka is incresibly good. Adding reference folders - duplicates inside those reference folders remain "unselected" and duplicates in any other folders are "selected" - will make Czkawka duplicate selection really useful. Currently, selecting the "correct" duplicate to delete is really limited. 

(Note: this reference folder functionality exists in software like dupeGuru.)

endolith:
This is different from https://github.com/qarmin/czkawka/issues/233 ?

nirvdrum:
Yes. This issue is about saying if you're searching multiple sources, only report duplicates if they are found in the referenced source. As @ca-boh noted, dupeGuru has similar functionality. #233 is about once those results are detected, allow me to show the duplicate list for only one of the sources.

I don't know if this helps, but say I'm searching for duplicates in X, Y, and Z. Currently, this project reports all duplicates in (X, Y), (X, Z), and (Y, Z) (and by extension (X, Y, Z)). The notion of a reference source would allow you to say only report duplicates if they appear in reference X, reducing the above list to (X, Y) and (X, Z).

Issue #233 would then allow you filter the list of results so you can only see the duplicates from X, Y, or Z. Given how easy it is to delete files from the application, such a filter would help prevent you from deleting files from the wrong source.

PR #516, opened 2021-12-22
Add reference folder

### LINKED ISSUE #313: File selection by original directory
The ability to mark all duplicates that have a candidate in /this/* directory
leaving all other duplicates unmarked.

When I start removing duplicates I always start by removing files that I already store in a sorted, more permanent places, like /music /pictures etc.

The ability to define these directories as containing _original versions_ would swiftly get the bulk out of the way leaving only duplicates that are all elsewhere to deal with.

### LINKED ISSUE #381: Is there a Database record?
Thanks Author for the program.

first time using czkawka, but would like to confirm below:

1. I have a folder with million of files, it will be great if this program can store all the files path and the size of each files, and possible the hash or hashmb of the files, so that when i add in new files i dont have to run through the full comparison again, possible?
I am aware it have cached support for faster in next scan, however i am not sure how long this cache will be kept as cache usually means volatile.

2. it is possible to set main folder and secondary folder? to allow user to always make comparison of secondary to main folder and remove file in secondary folder if match found?

### LINKED ISSUE #402: Duplicate search between, but not within, two distinct directories?
I'm trying to figure out how to configure `czkawka` to perform a duplicate search across two directories, but ignoring duplicates within the same directory tree.

The use case is a `source` folder of images that have previously been processed by scripts to rename them and sort them into a `destination` directory structure (e.g. by year and month, or by other EXIF metadata). Then we come across a new folder of images, some of which may have been processed previous.

1.  We'd like to know if any of those images already exist in the `destination` directory, presumably based on file contents.
2. The `source` folder may have duplicates in it, i.e. identical photos with different names. These are not of interest to be found (i.e. would like to ignore these duplicate matches).
3. It's possible that the `destination` folder also has duplicates in it, for example if you are storing multiple copies of a phone, one under a directory tree using a date-based organization, and one under a directory tree using EXIF-based organization (e.g. camera models). These duplicates may be real or soft/hard links. Either way, duplicates within the `destination` directory should be ignored.

So the only duplicates we're looking for are files in `source` that show up somewhere under `destination`.

Is this use case supported by `czkawka`?

### LINKED ISSUE #492: Is there a way to not scan for duplicates inside each respective folder that I added.
Sorry, if it's not appropriate to ask this in the issues.
I have a folder with files recovered from a hard drive that I accidentally wiped out and which I want to back up. And I also have a folder where I back up things. I want to check what I have already backed up that is present in the recovered folder in order not to waste space. Is there a way to do it? So like, I want czkawka to search for duplicates that are present in the folders A and B, but I don't want it to search if there are duplicates in the folder B itself.

qarmin:
Fix #381 (at least 2 point, 1 was fixed by adding prehash caching)
Fix https://github.com/qarmin/czkawka/issues/232
Fix #492
Fix #402
Fix #313

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,7 +11,7 @@
 - Option to not remove cache from non existent files(e.g. from unplugged pendrive) - [#472](https://github.com/qarmin/czkawka/pull/472)
 - Add multiple tooltips with helpful messages - [#472](https://github.com/qarmin/czkawka/pull/472)
 - Allow to cache prehash - [#477](https://github.com/qarmin/czkawka/pull/477)
-- Improve custom selecting of records(allows to use Rust red regex) - [#489](https://github.com/qarmin/czkawka/pull/478)
+- Improve custom selecting of records(allows to use Rust regex) - [#489](https://github.com/qarmin/czkawka/pull/478)
 - Remove support for finding zeroed files - [#461](https://github.com/qarmin/czkawka/pull/461)
 - Remove HashMB mode - [#476](https://github.com/qarmin/czkawka/pull/476)
 - Approximate comparison of music - [#483](https://github.com/qarmin/czkawka/pull/483)
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 - CLI frontend - for easy automation
 - GUI frontend - uses modern GTK 3 and looks similar to FSlint
 - No spying - Czkawka does not have access to the Internet, nor does it collect any user information or statistics
-- Multilingual - app support multiple languages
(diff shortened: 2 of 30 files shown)


felciano:
This sounds very tantalizing -- is there a description of that the Reference Folder notion represents formally, and how to use it to structure dedup tasks?

qarmin:
Well, I don't know if my implementation is correct and if I properly understand what exactly mean that there is set reference folder, but I tried to copy solution from dupeguru - https://dupeguru.voltaicideas.net/help/en/folders.html

PR #531, opened 2021-12-29
Implement DirTraversal

pczarn:
Uses a builder pattern to provide a common mechanism for file and directory search.

TODO: migrate other search categories.

### DIFF
--- a/czkawka_core/src/common_dir_traversal.rs
+++ b/czkawka_core/src/common_dir_traversal.rs
@@ -78,11 +78,18 @@ pub struct FolderEntry {
 
 // Collection mode (files / empty folders)
 
-#[derive(Eq, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq)]
 pub enum Collect {
     EmptyFolders,
     InvalidSymlinks,
     Files,
+}
+
+#[derive(Eq, PartialEq)]
+enum EntryType {
+    File,
+    Dir,
+    Symlink,
 }
 
 pub struct DirTraversalBuilder<'a, 'b, F> {
@@ -380,8 +387,18 @@ where
                                 continue 'dir;
                             }
                         };
+                        let entry_type;
                         if metadata.is_dir() {
-                            if collect == Collect::Files {
+                            entry_type = EntryType::Dir;
+                        } else if metadata.file_type().is_symlink() {
+                            entry_type = EntryType::Symlink;
+                        } else if metadata.is_file() {
+                            entry_type = EntryType::File;
+                        } else {
+                            unreachable!();
+                        }
+                        match (entry_type, collect) {
+                            (EntryType::Dir, Collect::Files) | (EntryType::Dir, Collect::InvalidSymlinks) => {
                                 if !recursive_search {
                                     continue 'dir;
                                 }
@@ -396,7 +413,8 @@ where
                                 }
 
                                 dir_result.push(next_folder);
-                            } else {
+                            }
+                            (EntryType::Dir, Collect::EmptyFolders) => {
                                 atomic_entry_counter.fetch_add(1, Ordering::Relaxed);
                                 let next_folder = current_folder.join(entry_data.file_name());
                                 if excluded_items.is_excluded(&next_folder) || directories.is_excluded(&next_folder) {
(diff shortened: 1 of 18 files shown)


pczarn:
@qarmin what do you think about the code, is this the right direction?

qarmin:
Removing duplicated code is always good, but I have some concerns about it
- performance - slightly lower performance at the expense of unification is a small price to pay but I don't benchmarked this yet(I think that checking by name or size may have the biggest impact in performance)
- cache size - the biggest concern I have about cache size. Entire FileEntry struct is used in Duplicate, Similar Images, Similar Videos and Broken Files to save data to cache, so struct as small as possible is required to be able to quicly load and save data to cache. Adding next fields like e.g. image hash(which is `Vec<>`) will increase size of cache. I don't have for now any numbers, so it is possible that cache size may increase only in e.g. 5-10% which is acceptable.

pczarn:
So I suppose after you will check performance, we can merge this PR. I can add memory statistics in another PR accoring to this post https://stackoverflow.com/questions/30869007/how-to-benchmark-memory-usage-of-a-function

pczarn:
@qarmin I believe small cache size can be retained with an implementation that is generic over additional file entry information. I can do this in a future PR. How about we divide the work over multiple PRs for easier code review?

qarmin:
Looks that PR is OK and I could merge it because I didn't find any performance degradation and also cache for now is same as before(only duplicates from changed modes in this PR uses cache system, but it is my custom implementation(similar image use bincode)).

The only problem is with invalid symlinks, because a little after start I have this crash
```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', czkawka_core/src/common_dir_traversal.rs:636:45
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/panicking.rs:100:14
   2: core::panicking::panic
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/panicking.rs:50:5
   3: czkawka_core::common_dir_traversal::set_as_not_empty_folder
   4: czkawka_core::common_dir_traversal::DirTraversal<F>::run
   5: czkawka_core::invalid_symlinks::InvalidSymlinks::find_invalid_links
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

```

pczarn:
@qarmin I could not reproduce the problem on my system. Could you send me test data? I would like to make a regression test case.

Even though I could not reproduce the problem, or figure it out by reading the code, I found bugs in my code. Please try symlinks again with the latest commit -- they should work.

czkawka_core/src/common_dir_traversal.rs
```diff
@@ -0,0 +1,664 @@
...
+                        };
+                        let entry_type;
+                        if metadata.is_dir() {
+                            entry_type = EntryType::Dir;
+                        } else if metadata.file_type().is_symlink() {
+                            entry_type = EntryType::Symlink;
+                        } else if metadata.is_file() {
+                            entry_type = EntryType::File;
+                        } else {
+                            unreachable!();
```
qarmin: This is reachable and should be ignored
```
thread '<unnamed>' panicked at 'internal error: entered unreachable code', czkawka_core/src/common_dir_traversal.rs:398:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
```
drwx------  2 rafal rafal 4096 sty  1 10:42 .
drwx------ 56 rafal rafal 4096 sty  1 07:43 ..
-rw-------  1 rafal rafal 1812 gru 11 09:25 Configuration.xml
prw-------  1 rafal rafal    0 sty  1 07:12 .show-request-queue <---------- This cause crash
```

qarmin (changes requested):

czkawka_core/src/common_dir_traversal.rs
pczarn: @qarmin fixed in the most recent commit

qarmin:
I tried several times to find symlink which cause crash with older version of this PR, but I couldn't find it and looks that only crash happens with specific combination of folders and files.

qarmin:
Thanks!

Issue #443, opened 2021-10-31
Select within groups : longest path or shortest path


willhubs:
I find that after doing a recovery, it's difficult to sort through some of the duplicates but I can easily control the length of the path.

If there was a selection choice that allowed you to choose all within groups except the shortest or longest path filename, it would make selecting them easier.

willhubs:
If czkawka always displays the shortest path first and the longest path last, then a 'select within groups all but first' and 'select within groups all but last' would work as well.

willhubs:
> If czkawka always displays the shortest path first and the longest path last, then a 'select within groups all but first' and 'select within groups all but last' would work as well.
After some testing, it doesn't always work out that way

ljhcage:
Indeed, sometimes longer filename which generated by system to avoid conflict is more accurately

willhubs:
An example is a bunch of random mismatched music files in:
/music

While the organized music is in:
/music/A Perfect Circle
/music/A Tribe Called Quest
/music/ABBA
/music/AC/DC
/music/AFI
/music/Aaron Carter
/music/Aaron Hall
/music/Aaron Tippin
/music/Ace of Base
/music/Adam Sandler
/music/Aerosmith

The random/unorganized ones would end up with the shortest path, not shortest filename, would end up being in /music 
Then I could choose> select within groups, all but the longest path, or 1 with the shortest path


Or the opposite, where I would prefer to keep the ones in /music and not the ones in /music/subdirectory
Then I could choose> select within groups all but the shortest path, or 1 with the longest path.

JoBoCl:
https://github.com/qarmin/czkawka/commit/f279e92257e29416fee551e3b564257226d8decc - this can likely be marked as complete

Issue #456, opened 2021-11-21
Moving large number of files just results in hanging


dalanmiller:
Firstly, _dziekuje ci_! This is such a helpful tool! I've been poorly trying to write a python script for a couple years to do the same thing but this is just so much better! 

I noticed that after scanning many thousands of files and attempting to move a slice of them to a new directory, after selecting 'Ok', the program just hangs. Looking at the directory I can see that files are indeed moving there but there's no indication that the program is still functioning or making progress or has hit some sort of issue. 

Perhaps a [progress bar](https://docs.gtk.org/gtk3/class.ProgressBar.html) status window would be best shown here so that it's clear what's happening? And/or a small window to indicate whether there are any issues with files as a outcome / results log? 

Just a couple ideas! 

<img width="1052" alt="Screen Shot 2021-11-21 at 8 54 41 pm" src="https://user-images.githubusercontent.com/863286/142757408-f88da952-5780-479d-8071-f97cce5f49ed.png">

bear-in-a-box:
I'd like to bump this one and provide some more details, as I've got the same issue.
On MacOS Monterey (12.0.1), something like this happened to appear in the terminal after trying to move items (selected all but the oldest ones):
```
ERROR: This should not happen model.value(&iter, column_color).get::<String>().unwrap() != MAIN_ROW_COLOR
thread 'main' panicked at 'First deleted element, should be a header', czkawka_gui/src/connect_button_move.rs:222:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
That first line appeared a few hundred times.

I'll be back with more details once I run the app with backtracing enabled.

gabefair:
I also noticed that it gets stuck at 

<img width="984" alt="Screen Shot 2022-10-03 at 1 33 50 PM" src="https://user-images.githubusercontent.com/2096785/193709834-127beff3-0486-4cdb-b23c-9049828f512d.png">

GeckoEidechse:
On Windows, running task manager while moving duplicate files I noticed a lot of disk usage and just moving 2 files with a total of 6GB took multiple minutes.

This would indicate that when moving a file, instead of simply updating the path where it is located, the old file is actually read into memory and written to a new file. Looking at the creation and last modified date of the moved file confirms that theory.

I'm not sure what the reason for this approach is, as unless one is moving files across disks/shares, simply updating the path the file is located at is faster many times over.

[...]

Issue #380, opened 2021-06-18
Limit CPU option


jose1711:
My CPU gets very hot when computing similarity of the images. I wish there was an option to limit number of cores (I know `nice` is another option) at the expense of longer waiting times.

necros2k7:
Option to choose number of CPU threads would be perfect

qarmin:
Such feature is already present in app

Issue #298, opened 2021-03-20
Duplicates "check method" should be checkboxes, not radio buttons


endolith:
![image](https://user-images.githubusercontent.com/58611/111880310-29ccb480-8981-11eb-96fe-b5aa8c2cf833.png)

I want to find duplicates that have the same name, same size, _and_ same file content.  I don't want it wasting time comparing two files that don't have the same name.

gessel:
I'd like to add a use case where this enhancement would be important - at least size+name: both size and name are likely to result in collisions in a large file set.  I found two different 131+mb video files with exactly the same file size, for example (unlikely but, it turns out, possible).   Had I been able to select both name and size I wouldn't have risked accidentally deleting a wanted record.

I'd argue hash is different - hash tests seem to imply an initial "size" check first.  

Counting on files having the same name seems an edge case in that many bulk copy tools (such as EXIF-sorter) will automatically rename files to prevent accidental overwrite, which can result in large amounts of accumulated file lint.

PatrickL546:
Any update on this? I'd like to find dupes by hash but I want to keep the dupe if it has a different name.
ex.
file1.ext = hash 1
file1.ext = hash 1

file2.ext = hash 1

delete the dupe of file1 but keep file2 even tho they have the same hash.

PR #538, opened 2022-01-01
[WIP] Implement sorting by portion checked within group

### LINKED ISSUE #444: Show next group that has all duplicates selected
Once in a while I will select from a path and will get some groups where all of the duplicates are selected.

A button, keyboard shortcut, or both would be great to take me to the next group that has all duplicates selected instead of going through the whole list manually.

pczarn:
How do I hide the button on startup?

Do you think this is a useful feature?

Fixes #444

### DIFF
--- a/czkawka_gui/src/compute_results.rs
+++ b/czkawka_gui/src/compute_results.rs
@@ -500,7 +500,7 @@ pub fn connect_compute_results(gui_data: &GuiData, glib_stop_receiver: Receiver<
                         set_specific_buttons_as_active(
                             &shared_buttons,
                             &NotebookMainEnum::Duplicate,
-                            &["save", "delete", "select", "symlink", "hardlink", "move"],
+                            &["save", "delete", "select", "symlink", "hardlink", "move", "sort"],
                             duplicates_number > 0,
                         );
 
@@ -565,7 +565,7 @@ pub fn connect_compute_results(gui_data: &GuiData, glib_stop_receiver: Receiver<
                         set_specific_buttons_as_active(
                             &shared_buttons,
                             &NotebookMainEnum::EmptyDirectories,
-                            &["save", "delete", "select", "move"],
+                            &["save", "delete", "select", "move", "sort"],
                             empty_folder_number > 0,
                         );
 
@@ -631,7 +631,7 @@ pub fn connect_compute_results(gui_data: &GuiData, glib_stop_receiver: Receiver<
                         set_specific_buttons_as_active(
                             &shared_buttons,
                             &NotebookMainEnum::EmptyFiles,
-                            &["save", "delete", "select", "move"],
+                            &["save", "delete", "select", "move", "sort"],
                             empty_files_number > 0,
                         );
 
@@ -699,7 +699,7 @@ pub fn connect_compute_results(gui_data: &GuiData, glib_stop_receiver: Receiver<
                         set_specific_buttons_as_active(
                             &shared_buttons,
                             &NotebookMainEnum::BigFiles,
-                            &["save", "delete", "select", "move"],
+                            &["save", "delete", "select", "move", "sort"],
                             biggest_files_number > 0,
                         );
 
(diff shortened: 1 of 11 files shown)


qarmin:
I'm not sure if I correctly understand what exactly functionality is needed.

In issue 422(not 444) user probably wants to manually sort results by path, name, size etc. column inside groups.
Looks that feature is the first the most wanted one by users, so I don't see reason to not to implement this.
This is already implemented for simple treeviews(without header row) like in Big files or Invalid Symlinks by built-in GTK sort feature(clicking at name of column will sort results).
https://github.com/qarmin/czkawka/blob/51df63b8eab993ecb32768a329b191b63e8afc4b/czkawka_gui/src/create_tree_view.rs#L215

Looking at source code from this PR I don't know what exactly it do.
When I tested it, then sometimes entire group of results jumped to the top of treeview, but usually nothing happens.

pczarn:
@qarmin The main part of the mechanism is:
```rust
list.push((num_selected as f64 / current_index as f64, range));

// ...

    list.sort_by(|&(portion_a, _), &(portion_b, _)| { portion_a.partial_cmp(&portion_b).unwrap() });
    let oldpos: Vec<_> = list.into_iter().flat_map(|(_, range)| range).map(|v| v as u32).collect();
    model.reorder(&oldpos[..]);
```

> When I tested it, then sometimes entire group of results jumped to the top of treeview, but usually nothing happens.

Groups of results with many checked entries will jump to the top or the bottom.

In #444 a user requested a way to go to the next group that has all duplicates selected.

The idea is that if such groups jump to the top of the list, the user can easily go through all such groups.

I can see that a good explanation may be necessary for users to understand this feature. How about a tooltip containing a description, would that be enough?

qarmin:
Hmmmm.... 
I always think about button which would move focus to group where all results are selected.
Current solution is quite non-intuitive for me, because moves elements around entire treeview and this will break current sorting in duplicates, where at the top are show the biggest files(hash, size) or with latest name in dictionary(name).

Also sort button is quite misleading, because users probably think that this will sort results by name, path etc.
This is not a problem if to this button will be added options to sort results in group by e.g. path

pczarn:
@qarmin ok, you are right about this feature. I will back down on this feature, also because if you didn't grasp it at first, users won't grasp it as well.

However, it may be useful to repurpose this PR to fix #443 by doing
* sort by path length within group
* + sort by filename length within group

and adding this plus

* select all except last
* select all except first

so that a user may do both actions (sort + select) and arrive at selected only longer path lengths or shorter path lengths, thus fixing issue #443.

For consistency, we may indeed add options to sort results in group by e.g. path, size:
* sort by path within group
* sort by size within group

Is there a need to make a distinction between sorting by filename and by path?

Also I would do a jump menu to cover #292 as well, so that there would be two kinds of jumps
* jump to the next group with all entries checked
* jump to the next selected entry

Sounds like a plan?

pczarn:
@qarmin you're the expert. Could you assign me some issues to work on? Also, since we both speak Polish it would be handy to exchange cell phone numbers (I won't post mine in public, but I can send you mine via reddit). This way, we could discuss whatever needs to be done, on the phone. Is that ok?

qarmin:
I don't have any strict plans what features/bugfixes should be available in next version, but I think that this issues could be resolved: 
- Built icons into repository(need to find icons with proper license and with which looks good(I like Yaru icons, but are GPL licensed)) - https://github.com/qarmin/czkawka/issues/588
- Add progress bar when moving/deleting files - https://github.com/qarmin/czkawka/issues/456
- Sorting results in groups
- Limit used CPU cores - https://github.com/qarmin/czkawka/issues/380 - only in similar images/videos because it can cause OOM in some situations(many cores and not too much RAM)
- Additional button in duplicate finder, to find only in Size and Hash check method, files with identical names - https://github.com/qarmin/czkawka/issues/298 (it should just before)
- Select/Unselect Custom by multiple criteria - https://github.com/qarmin/czkawka/issues/241
- Limiting column size -  https://github.com/qarmin/czkawka/issues/169

but of course any other features can be also merged if not broke too much current app workflow

micwoj92:
This branch has conflicts that must be resolved

Issue #556, opened 2022-01-04
Possibly a serious bug that need urgent attention


chchia:
@qarmin Please I would like to have your urgent attention, i found possibly a serious bug.

under same parameter, blockhash, 16 hashsize, similarity 5:

a. below without using fast compare in finding similar images, it show 32 result, and i check through it all appear to be different file. 
**this is correct**
![image](https://user-images.githubusercontent.com/1291771/148032176-de818e3d-bf35-4724-995f-413e74588a85.png)

b. **BUT** when using **WITH** fast compare enabled, it show 6824 result, and when i check details, it appear all group is identical files, if the user delete one of the files it means automatically delete both?? every group is same folder with same filename!
**this is wrong!!**
![image](https://user-images.githubusercontent.com/1291771/148030243-c77e2f39-fbe7-4fc5-a708-474965c7e0ad.png)
![image](https://user-images.githubusercontent.com/1291771/148030451-05e69ff3-6b11-4592-892e-4d2fb4a580b9.png)

while fast compare tooltip say
`        Speedup searching and comparing hashes.
        In opposite to normal mode where each hash is compared to each other x times, where x is similarity which user choose, in this mode always only one comparing is used.
        This option is recommended when comparing >10000 images with non 0(Very High) similarity.`

I believe this is possibly a very serious issues which will result in wrongly deleting many none duplicate files.

kindly please have a look when you have time, thank you 👍

chchia:
seems fixed, thank you for prompt action!

chchia:
@qarmin Hi qarmin, it seems latest fix did not fully resolve the problem, some file still duplicate in the list

that 2 files in red rectangle is in same folder same file name.
![image](https://user-images.githubusercontent.com/1291771/148147255-e78a14bc-1d54-46fc-bf57-836358f5153a.png)

PR #557, opened 2022-01-04
Fix duplicated entries when using fast compare option

qarmin:
Fixes https://github.com/qarmin/czkawka/issues/556


chchia:
Hi qarmin, I provide further update, the problem is not fully resolve.

https://github.com/qarmin/czkawka/issues/556

PR #558, opened 2022-01-04
Add cache support to similar music files

qarmin:
Also simplify cache code
Also add option to load json file when bin file is not available
Also add option to save cache results to json file


pczarn:
the title isn't right -- this is for videos and images

qarmin:
I not added yet cache support in first commit, for now I just simplified code.

PR #559, opened 2022-01-05
Use DirTraversal for similar music

pczarn:


pczarn:
@qarmin please check correctness on your test data -- I don't have any similar music on my system

qarmin:
Thanks!

PR #562, opened 2022-01-06
build with Docker

alexislefebvre:
I didn't wanted to install Rust on my system so I used Docker.

Thanks to these sources:
- https://github.com/jlesage/docker-czkawka/blob/master/Dockerfile
- https://stackoverflow.com/questions/49676490/when-installing-rust-toolchain-in-docker-bash-source-command-doesnt-work/58169817#58169817


qarmin:
Tested and works fine.
Thanks!

PR #572, opened 2022-01-10
Reverse selection when clicking at middle mouse button

qarmin:
Fixes https://github.com/qarmin/czkawka/issues/488
Fixes https://github.com/qarmin/czkawka/issues/465

This reverse selection of all items in group, except the one clicked(it will be more used than total reverse) e.g.

TreeView
HEADER
x - 1
x - 2
x - 3 <- clicked
x - 4
x - 5

later
HEADER
_ - 1
_ - 2
x - 3 <- clicked
_ - 4
_ - 5

PR #573, opened 2022-01-11
Document middle click

alexislefebvre:
See #572

### DIFF
--- a/instructions/Instruction.md
+++ b/instructions/Instruction.md
@@ -34,6 +34,7 @@ To open multiple file just select desired files with CTRL key pressed and still 
 
 To open folder containing selected file, just click twice on it with right mouse button.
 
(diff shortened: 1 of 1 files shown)


instructions/Instruction.md
```diff
@@ -34,6 +34,7 @@ To open multiple file just select desired files with CTRL key pressed and still
 
 To open folder containing selected file, just click twice on it with right mouse button.
 
+To invert a selection of files, click on a file with the middle mouse button.
```
qarmin: Here should be info that not all records are inverted(all except the clicked one)

alexisLefebvre: Thanks for the review, I added that.

qarmin:
Thanks!

PR #603, opened 2022-01-22
Fixing czech translation

habernal:
Just noticed a non-sense translation in the main GUI window.


qarmin:
Czech translation is computer generated, so such invalid messages are expected

Translations should be changed only in https://crowdin.com/project/czkawka, because it is easier to manage multiple languages at once with it

qarmin:
Thanks!

habernal:
OK, sorry, that wasn't clear to me from the readme.

alexislefebvre:
Yes [this page](https://github.com/qarmin/czkawka/blob/master/instructions/Translations.md#translations) mentions Crowdin and also explanations about the files, we may made it more clear that Crowdin must be used.

PR #604, opened 2022-01-22
Improve desktop entry

cyqsimon:
The word for `hiccup` is quite rude and/or vulgar in Chinese (at least in my opinion), so I didn't bother adding a translation for the `Name` field. It's common for many programs to not have a Chinese name anyways (e.g. Google Chrome).

`Keywords` field improves searchability in the launcher. The word `czkawka` is difficult to remember for people who don't speak Polish.


qarmin:
Thanks!

PR #607, opened 2022-01-23
Fix appstream file

TheEvilSkeleton:
This MR fixes the appstream file to fix the error produced in https://github.com/flathub/com.github.qarmin.czkawka/pull/11.

```
$ flatpak run org.freedesktop.appstream-glib validate data/com.github.qarmin.czkawka.metainfo.xml
data/com.github.qarmin.czkawka.metainfo.xml: OK
```


qarmin:
Thanks!

PR #623, opened 2022-02-14
Fix accumulated stop events

### LINKED ISSUE #622: GUI: Multiple stop messages can accumulate before a search is cancelled
The application can take a few seconds to respond after hitting the `Stop` button during a search and it's possible to click `Stop` more than once before the search is actually cancelled. 

Since the channel used for the stop messages is unbounded it's not possible to successfully search again until you hit the search button enough times to clear out those stale stop requests (the search just gets immediately canceled.)

KerfuffleV2:
Change stop message channel to bounded with size of 1 and ensure it's clear before starting a search operation.

Closes #622


qarmin:
Thanks!

I have seen problem before, but didn't find reason why this happens

PR #625, opened 2022-02-14
Periodically check for and handle pending GTK events while performing long operations.

### LINKED ISSUE #624: Linux/Wayland: The czkawka-gui can lose connection to the compositor during long operations
I don't know if this also applies to X or other OSes but I could consistently reproduce the problem by trying to dedup via symlink large amounts of files. The app would get greyed out due to it not responding to messages and then eventually crash with:

```
Gdk-Message: 02:41:44.084: Lost connection to Wayland compositor.
```

I believe the solution is to periodically check for GTK events. This does not affect searching, probably because the status window updates allow GTK to process events before they are delayed too long.

KerfuffleV2:
**Please note**: I really don't know much about GTK (especially with Rust) so this may or may not be the correct approach.

This change checks for and handles a pending GTK event once every 100 files processed. It may make sense to handle all pending events. The frequency of event checking could obviously be changed quote easily, I don't know what the optimal value is but 100 seemed reasonable.

Closes #624

### DIFF
--- a/czkawka_gui/src/connect_things/connect_button_delete.rs
+++ b/czkawka_gui/src/connect_things/connect_button_delete.rs
@@ -292,7 +292,8 @@ pub fn empty_folder_remover(
     let mut messages: String = "".to_string();
 
     // Must be deleted from end to start, because when deleting entries, TreePath(and also TreeIter) will points to invalid data
-    for tree_path in selected_rows.iter().rev() {
+    for (counter, tree_path) in selected_rows.iter().rev().enumerate() {
+        handle_gtk_pending_event_counter(counter);
         let iter = model.iter(tree_path).unwrap();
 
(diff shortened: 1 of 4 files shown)


KerfuffleV2:
Any opinion on this? Maybe this particular pull isn't the correct approach but delaying events for an extended time isn't good and it definitely causes problems when using czkawka on Wayland.

qarmin:
I think that correct approach is to move deleting to different thread/threads.
I'm leaving this PR open, because I don't know if I will have enough time to implement thread support for move/delete operations until 4.1 version will come, so I treat this like backup solution.

I'm

qarmin:
For now is OK, but solution with moving computation to thread will fix also https://github.com/qarmin/czkawka/issues/456
Thanks!

PR #657, opened 2022-04-02
Add button to disable by default size of letters when selecting things

qarmin:
Add option to Ignore size of letters when selecting records


czkawka_gui/i18n/en/czkawka_gui.ftl
```diff
@@ -182,6 +182,11 @@ popover_custom_regex_check_button_entry_tooltip =
 
         This uses the default Rust regex implementation. You can read more about it here: https://docs.rs/regex.
 
+popover_custom_ignore_size_check_button_tooltip =
+        Ignore size of letters.
```
alexisLefebvre: I think it should be "case" instead of "size": https://en.wikipedia.org/wiki/Case_sensitivity

Issue #373, opened 2021-06-04
Cannot symlink or hardlink: can only delete dupes


Rings4:
Scanned 2 directories, clicked "select all except oldest", then clicked hardlink: nothing happened. Symlink didn't do anything either. Delete works, but I don't want to delete.

NAHNAH55:
Same issue. Symlink function does nothing. I wish I knew this before spending all day scanning my files.

qarmin:
What OS?
If windows, then for now only simlinks are implemented.
For unix like os(which use inode like Linux or macOS) it should work both symlink and hardlink

NAHNAH55:
Win 10. I haven't tried hard links, but symlinks don't work. I've tried both elevated and not elevated access. I select an option, press the symlink button and the files are deselected. No change takes place, neither file has been replaced with a link.

Rings4:
dupeGuru works with symlinks and hardlinks just fine. Perhaps you can look at what they're doing to implement into your own software? https://dupeguru.voltaicideas.net/

qarmin:
The problem is mostly with functions which Rust standard library provide(Rust and Python differs a lot)
Symlink on Windows problem should be fixed by https://github.com/qarmin/czkawka/pull/409

qarmin:
Hard-links should probably works with the latest version(future 3.2 version)

mascherone108:
I'm using Kubuntu 21.04 and it still doesn't work for me just as OP described.

qarmin:
Oh, now I understand what is happening(I think that this behavior is somewhere documented).

Symlinks and Hardlinks can be generated only if in group is at least 2 selected results.
1 result is always unchanged and second and furthers are hardlinked to first.

I reopened issue because in future I probably can implement this(this require to do some deep changes in method how hardlinks or symlinks are handled)

Bartek200219:
When I try to create hard link program crashes.
(Czkawka 3.2.0
Windows 10 2004)

qarmin:
Crash when hardlinking is fixed in master branch, but still happens with 3.3.1 version and it show when clicking at hardlink/symlink button without any record selected.

Warning message about it is added in this PR - https://github.com/qarmin/czkawka/pull/489

For now I don't see simple and clear for user method to set target of file used in hardlinking

z0diaq:
What about drag and drop within groups to change ordering ?

z0diaq:
[...]

PR #668, opened 2022-04-06
No safe hardlinking for Windows

qarmin:
Fixes https://github.com/qarmin/czkawka/issues/373 (I think)

### DIFF
--- a/czkawka_core/src/duplicate.rs
+++ b/czkawka_core/src/duplicate.rs
@@ -1248,20 +1248,15 @@ fn filter_hard_links(vec_file_entry: &[FileEntry]) -> Vec<FileEntry> {
 }
 
 pub fn make_hard_link(src: &Path, dst: &Path) -> io::Result<()> {
-    dbg!("1");
     let dst_dir = dst.parent().ok_or_else(|| Error::new(ErrorKind::Other, "No parent"))?;
-    dbg!("2");
+    dbg!("1 -  {}", dst_dir);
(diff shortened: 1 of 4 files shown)


czkawka_core/src/duplicate.rs
```diff
@@ -1250,12 +1250,12 @@ fn filter_hard_links(vec_file_entry: &[FileEntry]) -> Vec<FileEntry> {
 pub fn make_hard_link(src: &Path, dst: &Path) -> io::Result<()> {
     let dst_dir = dst.parent().ok_or_else(|| Error::new(ErrorKind::Other, "No parent"))?;
     dbg!("1 -  {}", dst_dir);
-    let temp = tempfile::Builder::new().tempfile_in(dst_dir)?;
-    dbg!("Trying to copy {} --- {} --- {}", &src, &dst, temp.path());
-    fs::rename(dst, temp.path())?;
+    let temp = dst_dir.join("rzeczek.rxrxrxl");
```
alexisLefebvre: @qarmin What is there an animal here? :slightly_smiling_face: [rzeczek](https://pl.wikipedia.org/wiki/Rz%C4%99sorek_rzeczek)

qarmin: I created rzeczek as the incorrect masculine form of the noun rzeczka(small river).
I thought such a meaningless name did not exist, but I was wrong ¯\\\_(ツ)\_/¯.
But still, combined with the extension there is almost 0% chance that anyone will ever use it, so it is good as is.

PR #673, opened 2022-04-12
Add exclude other filesystems option to cli

### LINKED ISSUE #31: Add a `find -xdev` equivalent
Add a `find -xdev` equivalent. Add a function to only search on one filesystem.

ongchi:
Add `-X, --exclude-other-filesystems` to exclude files on other filesystems then include directories

Fixes #31


qarmin:
Thanks!

PR #674, opened 2022-04-12
Misc improvements

Gitoffthelawn:
I haven't tried this tool yet, but it looks great!

While reading the README, I found some errors, so I fixed them.

There is one sentence that is still unclear because I'm not sure what you meant:
*"I prepared a disk and performed a test without any folder exceptions and with disabled ignoring of hard links."*

Did you mean that you specified to ignore hard links or that you specified to follow hard links?


qarmin:
Ignored hardlinks means in this app, that only 1 file from group of hardlinks is visible in results(it is enabled by default).

qarmin:
Thanks!

Issue #611, opened 2022-01-26
Music Duplicates doesn't search AAC(.m4a) files anymore


Jona-D:
Using the latest version 4.0.0 the Music Duplicates function no longer searches through iTunes ACC(.m4a) files. If I do the same search using version 3.3.1 .m4a files do show up in the search results.

Issue #651, opened 2022-03-20
[Feature Request] Broken Files - Add option to rename files with incorrect file extension


dnutan:
I think it would be a nice feature to be able to correct the file extension on those Broken Files where the file extension does not match the file header signature / magic number (not 100% accurate due to some file formats using container formats).
This could work for images, .CHK files, etc. either applying a suggested file extension to all selected files or with a custom file extension (on GUI an extra editable column with suggested extension).

**About File Signatures**
* https://en.wikipedia.org/wiki/List_of_file_signatures
* https://www.garykessler.net/library/file_sigs.html
* https://www.nationalarchives.gov.uk/PRONOM/Default.aspx
* https://filesignatures.net/index.php?page=all  (webpage with unsecure certificate warning)

**Some programs that can detect file signatures**
* [TrID](https://mark0.net/soft-trid-e.html)
* `file` command (based on libmagic library) (GNU/Linux)
* `exiftool` (for images, can also extract metadata useful to rename filename by date or other fields)
* [DROID](https://github.com/digital-preservation/droid)
* Windows PowerShell  (`Get-FileMetaData -Signature`)
* File Carving and File Recovery software 

- https://tika.apache.org/2.3.0/detection.html

---
Feature Request slightly related to #309

qarmin:
Most tools are written in C/C++ and similar, so integration can be problematic.

From Rust side, probably something like https://crates.io/crates/mime_guess can be used

dnutan:
For Rust, maybe something like [mime2ext](https://crates.io/crates/mime2ext), [infer](https://crates.io/crates/infer) (or [tree_magic](https://crates.io/crates/tree_magic) or [tree_magic_mini](https://crates.io/crates/tree_magic_mini)) would help doing the job.
Regards.

PR #678, opened 2022-04-17
Finding invalid extensions

qarmin:
Support for finding files with invalid extensions

Algorithm
- Get original extension of file
- Check what content is inside file(returns single extension)
- To provided extension, mime type is provided
- All extensions from this mime type are loaded
- App check if original extension is inside provided ones by mime  type

Due to not 100% perfect extension finding by used libraries, there are provided several workarounds to lower amount of not correctly detected records.

App sometimes can show invalid results especially with system files or less known extensions, so I suggest to use it only for well known own files like e.g. photo collections. 

TODO
Button to rename files(not sure how it should work, because most of files have multiple possible proper extensions)

Fixes https://github.com/qarmin/czkawka/issues/611
Related to https://github.com/qarmin/czkawka/issues/651


qarmin:
HELP IN FINDING PROPER LIBRARY NEEDED

Well,
Entire logic is inside, but I didn't read properly description of mime_guess crate -  "MIME/MediaType guessing by file **extension.**" XD (it was quite strange that ~500000 checked files had proper extensions)

So, I'm looking for replacement that library.
Currently looks that this library can be proper:
- https://crates.io/crates/content_inspector
- https://crates.io/crates/xdg-mime

Issue #699, opened 2022-04-29
Music Duplicates search gets stuck on "Reading Tags" phase


sleepwakecycle:
I'm doing an Approximate Comparison search by Title, Artist, and Length. Each time, a few seconds after starting the search, it gets stuck at 98% through while reading the tags. I've let it run for upwards of 30 minutes with no change. The "Stop" button doesn't work, I have to end it through task manager. I've tried this several times, with the same result.

qarmin:
I think that app crashes in one thread when reading tags from file.
Can you check which exactly files causes this?

sleepwakecycle:
How do I check what file causes the crash? I've narrowed it down to one folder, but the folder contains 1,831 files, so I'm not sure which one it is.

qarmin:
The fastest way is copying half of checked folders to another and checking if this one causes problems.

sleepwakecycle:
Ok, I found the 2 files that were causing the crash. Is there a way I can fix the files, so they can be read by Czkawka without crashing?

qarmin:
The only thing that you can do is to ignore this files, by avoiding scanning them or just directly pointing at them in Excluced Items field
Can you post this files here, to be able to check what the problem is?

sleepwakecycle:
I can't post them, they're M4As and GitHub's giving me the error message "We don't support that file type." I'll just manually exclude them. Thank you!

qarmin:
File can be zipped and then send via github.

With example file I could fix this bug, so excluding items won't be needed anymore

sleepwakecycle:
Ok, here are the files. Thanks again

[Czkawka glitch files.zip](https://github.com/qarmin/czkawka/files/8592423/Czkawka.glitch.files.zip)

Issue #696, opened 2022-04-27
Music Duplicates - length field isn't populated if length is not selected as a matching option


kaysond:
Running 4.1.0 on Windows Server 2k19. If I don't select length as a matching option, the length isn't shown in the results. It would be nice to have that as a means of comparison

![image](https://user-images.githubusercontent.com/1147328/165638452-03056f6e-79fa-4fa8-b7af-29802b552eb0.png)

qarmin:
If length or year is not visible, that means that music file doesn't have such info in their tags.

kaysond:
> If length or year is not visible, that means that music file doesn't have such info in their tags.

Which tags/frames are you looking for? Here is the media info dump for the files in the bottom of that screenshot. It has both year and duration.
```
Complete name                            : 03 - (defun botsbuildbots () (botsbuildbots)).mp3
Format                                   : MPEG Audio
File size                                : 9.49 MiB
Duration                                 : 4 min 7 s
Overall bit rate mode                    : Constant
Overall bit rate                         : 320 kb/s
Album                                    : Portal 2: Songs to Test By, Volume 2
Album/Performer                          : Aperture Science Psychoacoustics Laboratory
Album/Performer/Sort                     : Aperture Science Psychoacoustics Laboratory
Part/Position                            : 1
Part/Total                               : 1
Track name                               : (defun botsbuildbots () (botsbuildbots))
Track name/Position                      : 3
Track name/Total                         : 18
Performer                                : Aperture Science Psychoacoustics Laboratory
Performer/Sorted by                      : Aperture Science Psychoacoustics Laboratory
Composer                                 : Mike Morasky
Composer/Sorted by                       : Morasky, Mike
Publisher                                : Valve
Genre                                    : Electronic
Original/Released date                   : 2011
Recorded date                            : 2011-07-01
ISRC                                     : USA561361567
Cover                                    : Yes
Cover type                               : Cover (front)
Cover MIME                               : image/jpeg
SCRIPT                                   : Latn
Media Type                               : Digital Media
originalyear                             : 2011
MusicBrainz Album Status                 : official
MusicBrainz Album Release Country        : XW
MusicBrainz Album Type                   : album; soundtrack
WORK                                     : (defun botsbuildbots () (botsbuildbots))
ARTISTS                                  : Aperture Science Psychoacoustics Laboratory
MusicBrainz Work Id                      : c725dc05-e937-4eb5-aac4-cb03bd9993d7
[...]

Issue #690, opened 2022-04-24
Marks azw3 as bad extension


TheMetalCenter:
This may be working as intended since I believe azw3 is just an upgraded mobi file, but I wanted to point it out in case it wasn't. Every azw3 is marked as proper extension "mobi,prc". 

It would be nice if we could filter out certain extensions if this is intended.

Great app!

PR #705, opened 2022-05-01
Update ImagePipe

qarmin:
I wanted also update Bincode, since updating it is not so trivial, but for now this bug prevents from doing this - https://github.com/bincode-org/bincode/issues/537

Fixes: https://github.com/qarmin/czkawka/issues/696
Fixes: https://github.com/qarmin/czkawka/issues/690 - reported https://github.com/abonander/mime_guess/issues/72
Fixes: https://github.com/qarmin/czkawka/issues/699

Fixes bug when audio with lower than 1s length was visible without length(now is 0:01).
Also shows in parentheses which exactly extensions was found by bad extensions tool. 
Changed also Rodio to my own crate to open and check audio files(a lot of dependencies was deleted)

Added support for finding invalid PDF files(new unreleased version will provide a lot of more complete support)

Also updated minimal Rust version to 1.60 which is not yet available in Ubuntu, but it will be someday.

Waiting(in next PR) for updating:
Lofty-rs 0.6.3 - parses more files(unreleased and have more regressions)
Waiting for https://github.com/pdeljanov/Symphonia/issues/122, but for now I handle crashes in normal builds, so it is not so important, because for now it will just print  to output info about not proper data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ You can help by creating:
 - Bug reports - memory leaks, unexpected behavior, crashes
 - Feature proposals - proposal to change/add/delete some features
 - Pull Requests - implementing a new feature yourself or fixing bugs.
-  If the change is bigger, then it's a good idea to open a new issue to discuss changes.
+  If the change is bigger, then it's a good idea to open a new issue to discuss changes, but issues with label `PR welcome` are already checked and accepted to implement.
 - Documentation - There is an [instruction](instructions/Instruction.md) which you can improve.
 - Translations - Instruction how to translate files is available [here](instructions/Translations.md)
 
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -138,7 +138,7 @@ You can help by creating:
 - Bug reports - memory leaks, unexpected behavior, crashes
 - Feature proposals - proposal to change/add/delete some features
 - Pull Requests - implementing a new feature yourself or fixing bugs.
-  If the change is bigger, then it's a good idea to open a new issue to discuss changes.
+  If the change is bigger, then it's a good idea to open a new issue to discuss changes, but issues with label `PR welcome` are already checked and accepted to implement.
 - Documentation - There is an [instruction](instructions/Instruction.md) which you can improve.
 - Translations - Instruction how to translate files is available [here](instructions/Translations.md)
```
alexisLefebvre, line 144: The end of text is missing.

README.md
```diff
@@ -178,5 +178,7 @@ Big thanks to Pádraig Brady, creator of fantastic FSlint, because without his w
 
 Thanks also to all the people who create patches for this program, make it available on other systems, create videos, articles about it etc.
 
+Also I really appreciate work of people that create crates on which Czkawka basing and for that I try to report bugs to make it even better.
```
alexisLefebvre: "on which Czkawka is based"?

README.md
alexisLefebvre: My bad, I think it was a display issue with the GitHub app.

Issue #761, opened 2022-06-20
is Fast Compare no longer working?


chchia:
I clone the latest source and compiled, finding Similar Images with Fast Compare option on, yet still it comparing hash very slowly compared to previous version.

![image](https://user-images.githubusercontent.com/1291771/174521944-ce398ffb-8fbc-435e-968a-99aa5b25c1d6.png)

comparing speed is about 200 images per second....

chchia:
![image](https://user-images.githubusercontent.com/1291771/174523870-7a8fa107-17bf-4070-938e-138c4bc59081.png)
also, clicked stop and it show Stop requested, yet is still continue to run

qarmin:
Fast compare option has not changed since previous version

Also stop window works exactly same, but not when user click at button, there is shown message.

Currently only rarely code checks if user stopped search, because otherwise this will cause even lower check performance

qarmin:
Message is not clear enough

Comparing `x/y` image hash means that up to this moment `x` comparisons have been made of all `y`.

So having 1 million files not means that there will be 1 million comparisons(usually it is a lot of more).
Number of comparisons depends on number of group of images with same hash.

Issue #679, opened 2022-04-18
Low CPU utilization during image hash comparison


gotr3k:
Hi, was just wondering why is the CPU utilization low during image hash comparison phase of the scan? Every other phase has some obvious bottleneck (disk/cpu at max), but this one confuses me. How is the comparison implemented? Great work btw!

qarmin:
For now the main bottleneck is that algorithm works only on one thread 

Code - https://github.com/qarmin/czkawka/blob/c88d347e0093ba0af47b2fc8714e67cced36cc89/czkawka_core/src/similar_images.rs#L718-L836

Adding multithreading to it is not so easy, because finding objects is related to list of excluded items, that was modified in previous iterations, so simple and fast solutions of dividing tasks to 4,8 or 16 parts are not available here.

But even with support of multiple threads, it would be still quite slow because with big number of hashes, each hash needs to be compared with every else multiple times.

Issue #512, opened 2021-12-20
Stuck at hashing image stage


Erothas:
**Operating system**
Linux
**GUI or CLI**
GUI.. I think?
**Mode in question**
Similar image mode

![window-Mon_20_Dec_12 30 56_CET_2021](https://user-images.githubusercontent.com/55455670/146830226-c3e38298-3877-4c0f-9fa4-6bd35d799ebe.png)

As you can see above, I am stuck at this hashing image stage. I have left the program running for over 11 hours at one point, but it's perpetually stuck. I am trying to get down to what exactly is causing this through the process of elimination, but it's very difficult to put my hands on what exactly is the cause. A few things that may or may not influence this:

- My image sample size is admittedly gigantic. It's probably over half a million by now.
- ~~Higher hash size~~ Lowering it to the default 8 did nothing.
- Similarity. It seems like the lower the similarity, the more the problem occurs?

I don't even know what more information I can give you. I hope this is enough.

qarmin:
Can you run app from terminal?
Crash should print some info to console.
I suspect that one of images are broken and cause thread panic, because such situation is not handled in `image-rs` crate which I use.

Erothas:
![DP-2-Tue_21_Dec_18 26 22_CET_2021](https://user-images.githubusercontent.com/55455670/146973266-56659078-b50f-45d3-b1f0-7862da07561e.png)
As you can see, there is not much shown on the terminal. This has been running for close to 8 hours by now. Did I miss a command flag when starting the program or something?

qarmin:
By default simple `./czkawka_gui`  should print basic info about crash.
`RUST_BACKTRACE=1 ./czkawka_gui` would print more info, but only if app crashes, so in this situation it is useless.

Maybe helpful would be log from gdb, but this would require to execute this commands:
- `gdb`
- `file czkawka_gui`
- `run`
- Gui should open, so now required is to freeze app
- few second after freeze CTRL + C needs to be clicked
- `backtrace`
Something like this should print
```
#0  0x00007ffff6f52cdf in __GI___poll (fds=0x555555ca93d0, nfds=3, timeout=229) at ../sysdeps/unix/sysv/linux/poll.c:29
#1  0x00007ffff7205ea6 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007ffff71b0003 in g_main_context_iteration () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
[...]

PR #762, opened 2022-06-26
Change image hash compare algorithm and add multithreading

qarmin:
Closes https://github.com/qarmin/czkawka/issues/679
Closes https://github.com/qarmin/czkawka/issues/761
Helps https://github.com/qarmin/czkawka/issues/512

Algorithm performance for all similarity options should be very similar to fast compare option which become unnecessary and was removed.

Results should be a little worse than with previous algorithm, but scan speed boost especially with bigger similarities, should be clearly visible

Due using multi-threading, checking should be a lot of faster with bigger amount of files(0 similarity is special situation in which comparing images should happen almost instant), since most of operations are calculated on memory/CPU(without any disk operations)

Due removing Similarity enum, image cache file is not longer backward compatible, but still json config can be exported, changed manually and later imported to app(read docs for more info)

The only problem that I see is that I can't really test performance, because with ~8K images which I'm able to scan

With my testing on 20K images, new algorithm(4/8 processor) is a lot of faster than normal scan, but slightly less performant than fast compare
```
--- Similarity 20

New Normal 9
Old Fast Compare 6
Old Normal 106

--- Similarity 2

New Normal 2
Old Fast Compare 2
Old Normal 2
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,9 +68,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.57"
+version = "1.0.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08f9b8508dccb7687a1d6c4ce66b2b0ecef467c94667de27d8d7fe1f8d2a9cdc"
+checksum = "bb07d2053ccdbe10e2af2995a2f116c1330396493dc1269f6a91d0ae82e19704"
 
 [[package]]
 name = "arrayref"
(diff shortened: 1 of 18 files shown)


chchia:
> Due removing Similarity enum, image cache file is not longer backward compatible

that means i will have to rehash a total 1.8millions images again, that will be too bad, could you remain the option to use the old method? it will be too much for me to rehash 1.8 millions files as i store my file in online storage....

qarmin:
From the app perspective, using integer instead enum is only cosmetic change(enum with 1 field doesn't have much sense), but maybe this is not worth breaking changes

chchia:
i was thinking maybe you can consider within folder comparison.

example the user may have hundreds of folders, but the picture in each folder was actually categorize, example i might shooting the same scene for 30 photos with very minor different, and all those folder is put under a single folder.

so if there is a option let say "Compare images within the same folder only", with this i believe we can get a significant merit of:

1. shorter comparison time as we dont have to compare 1 file to all folder, but instead only image within the same folder.

2. with this, i can set the Similarity to even like 20, those same scene picture are more likely to be categorize as same. otherwise if comparing all photo it is more likely to see mistake.
image

PR #796, opened 2022-07-31
Added Korean localization to GUI

alexdraconian:


qarmin:
Thanks,
Now I manually imported translation to https://crowdin.com/project/czkawka, so any further improvements should be added there, not via commits on Github.

Issue #798, opened 2022-08-01
Similar Image ignore images that have same size with similarity >0.


chchia:
I found that Similar Image did not show duplicate if file size is same and similarity >0.

this can be simply reproduced, just copy the images for few copies inside the same folder. and Similar Image will not find anything.

when set similarity to 0 then it will find again.

PR #799, opened 2022-08-01
Attempt to fix similar images algorithm

qarmin:
Fixes https://github.com/qarmin/czkawka/issues/798


qarmin:
Looks that I finally found cause of this problem, so fix should come probably tomorrow

Issue #835, opened 2022-09-18
Bad performance when scanning large datasets on HDDs


sergeyvolk:
I have noticed that Czkawka performance can be suboptimal when scanning for dups on HDDs. As far as I can see from the source code you are using rayon::par_iter to read/scan several files simultaneously (for example  https://github.com/qarmin/czkawka/blob/e731f5ed75d9f4a2a04f183537b2fa7d2abfbce4/czkawka_core/src/duplicate.rs#L751). This works great for SSDs, but causes terrible performance for HDDs. That's because when you are reading multiple files simultaneously from an HDD where those files are located in different parts of the disk, a lot of time is wasted for seeking (moving mechanical read/write heads back and forth), in my case I observed that an external 4TB HDD which is typically capable of ~100-120MB/sec sequential read speed when reading a single file, was getting ~10-20MB/sec actual read speeds due to this excessive seeking, because czkawka was trying to read 4 files from it simultaneously (disk queue size observed in Windows Performance Monitor was around 4).

In order to get optimal performance for HDDs it would be nice to have an option to limit the number of threads used for scanning. Ideally it should be limited to 1 thread per physical HDD, but I understand that that's non-trivial to do without major refactoring. So in order to provide a quick workaround perhaps we could provide at least on option to limit the number of threads to 1 globally? Even that would be much better than trying to read 4 files from an HDD at once (and it only gets worse if you have a more powerful CPU with more cores). I can see in Rayon FAQ that it's possible to limit the number of threads by setting RAYON_NUM_THREADS env variable, but I'm using Windows GUI and don't know how to set that. Can we add a GUI option to set the number of threads or disable parallelism (i.e. not use par_iter() at all)?

qarmin:
`RAYON_NUM_THREADS=2 ./czkawka_gui.exe` from powershell should work(probably)

I tried to use allow user to change number of threads, but looks that this is allowed only once(later error is printed, that threadpool is already initialized) - https://github.com/qarmin/czkawka/pull/839

For now I have no idea how I could allow to run max 1 thread on HDD with no performance hit(probably each file should be marked with two variables `is_hdd` and `disc_id` to be able to set lock and release it after successful scan)

PR #839, opened 2022-09-24
Allow to set number of used threads in scan

### LINKED ISSUE #823: 5.0.2 unsupported on Debian Bullseye (due to libc >= 2.32 dependency)
Would it be possible to support 2.31? See output:

```
root@dunder:~# /root/czkawka dup -d /tank --delete-method HARD --file-to-save /root/czkawka_out.txt
/root/czkawka: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /root/czkawka)
/root/czkawka: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /root/czkawka)
/root/czkawka: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /root/czkawka)
root@dunder:~# ldd --version
ldd (Debian GLIBC 2.31-13+deb11u3) 2.31
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
root@dunder:~# lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 11 (bullseye)
Release:        11
Codename:       bullseye
root@dunder:~#
```

qarmin:
Part of #835
Fixes #823 - Automatically CLI is build in Linux with Ubuntu 18.04

Thread number can only be set once, so this is done at the start of the app, and settings are read from settings file

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -16,14 +16,14 @@ jobs:
         type: [ release ]
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: ${{ matrix.toolchain }}
           override: true
 
-      - uses: actions/cache@v2
+      - uses: actions/cache@v3
         with:
           path: |
             target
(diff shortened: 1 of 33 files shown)


Paruccone:
cool thx, but how do we set the # of threads from linux CLI? I tried to add env variable RAYON_NUM_THREADS as mentioned in #835 for windows but with no effect :( Went through help and no mention of it

Issue #841, opened 2022-09-27
Error in German translation


gimplyworxs:
Duplicate Files should be transalted as 
"Doppelte Dateien suchen" or "Finde Duplikate"
but definitely not as "Dateien duplizieren".
This is really misleading.

qarmin:
Changes to translations can be done in Crowdin

PR #843, opened 2022-09-30
Fixes and improves German translation

m-terlinde:
This is a fix for #841


m-terlinde:
I've just seen, that you use Crowdin. Shall I transport the changes to that platform?

qarmin:
Yes, this should changed in Crowdin(already someone changed this keys)

Issue #833, opened 2022-09-16
Windows 10 hardlinking duplicates is a bust.


Erquint:
Windows 10.
Found a group of two identical files by XXH3 residing on the same disk.
Selected both files in the group.
Pressed the Hardlink button. Proceeded with the confirmation prompt.
The group vanished from the list.
No new messages in the log.
None of the corresponding files in the filesystem gotten replaced by a hardlink. They remain redundant.

I installed and ran this software overnight on a large disk overnight in order to deduplicate files between two big backup archives on it. That wasn't the only group. There's more than I could hardlink manually. This has been a waste.

If hardlinking wasn't implemented on Windows — the readme should've warned me in advance instead of advertizing:
> __works__ on Linux, __Windows__, macOS, FreeBSD and many more

qarmin:
Last time it works when I checked this.
Windows require to have admin rights to do sym/hard links - http://security.stackexchange.com/questions/10194/ddg#10198 - so I only can add this info to documentation(of course if there is no regression)

Erquint:
Do you care to check now?

And how should the logging issue be resolved?
It would help to troubleshoot the issue if the log gave any information upon this action.

Erquint:
Thank you. Can't test yet, but hopefully this fixes it.

PR #881, opened 2022-12-20
Windows hardlinking problem

qarmin:
Fixes https://github.com/qarmin/czkawka/issues/833


Erquint:
I was reviewing the changes and couldn't help but notice that the English localization crept into localization files for other languages.
I don't know if that is intended as part of your workflow.

PR #882, opened 2022-12-20
Fixed mistranslated german label

andreasmuelhaupt:


qarmin:
Thanks, but this needs to be changed here - https://crowdin.com/project/czkawka

PR #895, opened 2023-01-09
Improve image preview quality

darktorres:
As demonstrated in the link: [https://docs.rs/image/latest/image/imageops/enum.FilterType.html](url)

### DIFF
--- a/czkawka_gui/src/connect_things/connect_button_compare.rs
+++ b/czkawka_gui/src/connect_things/connect_button_compare.rs
@@ -414,14 +414,14 @@ fn generate_cache_for_results(vector_with_path: Vec<(String, String, TreePath)>)
 
         #[allow(clippy::never_loop)]
         loop {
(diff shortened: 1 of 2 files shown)


alexislefebvre:
Bilinear doesn't appear on the link or when searching: https://docs.rs/image/latest/image/imageops/enum.FilterType.html?search=Bilinear

How does it work?

darktorres:
The manual for the change I made is in here:
[https://docs.gtk.org/gdk-pixbuf/enum.InterpType.html](https://docs.gtk.org/gdk-pixbuf/enum.InterpType.html)
The first link is only for the images examples.

darktorres:
Gdk Bilinear is closest to imageops Triangle

qarmin:
What is time difference in loading big images?
The main reason why I choose such resizing algorithm is quite big loading/resizing times of files with >10 millions pixels.
If such times are too big, then only smallest files should be used to resized with this algorithm

alexislefebvre:
@ImportTaste With this PR, it should be easier to compare similar images since there will be less artifacts in the previews of images.

PR #921, opened 2023-02-25
Remove preview workaround in gui for webp/heif

qarmin:
This adds new dependencies - `heif-gdk-pixbuf` and `webp-pixbuf-loader` which are used to show previews in gui(someday, this should be integral part of gdk-pixbuf)

This speedup loading preview of files with such extensions.

I thought it would be possible to get rid of libheif completely, but it turns out that it is still used when checking similar images

The only blocker is now missing heif-pixbuf in homebrew repo - https://github.com/strukturag/libheif/blob/master/gdk-pixbuf/pixbufloader-heif.c


qarmin:
webp already was merged, heif file cannot be added due missing packages in repo

PR #945, opened 2023-03-15
Update Installation.md for Mac M1/M2

aioue:
Add Apple Silicon installation.

### DIFF
--- a/instructions/Installation.md
+++ b/instructions/Installation.md
@@ -47,7 +47,19 @@ Prebuilt binaries are available only for x86_64, so if you use ARM e.g. Mac M1/M
 
 **Compilation is simple, and takes less than 10 minutes on an M1/M2 mac.**
 
-User [@bauchdj](https://github.com/bauchdj) wrote [working rust complication instuctions](https://github.com/qarmin/czkawka/issues/689#issuecomment-1310549801)
+```
+# Install Rust and Libraries
+brew install rustup
+rustup-init
+brew install gtk4 adwaita-icon-theme librsvg libheif
+```
+```
+# Clone and Build
+git clone https://github.com/qarmin/czkawka.git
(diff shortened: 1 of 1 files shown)


instructions/Installation.md
```diff
@@ -42,16 +42,21 @@ At the end execute it:
 ```shell
 ./mac_czkawka_gui
 ```
+#### Compiling on Apple Silicon
+Prebuilt binaries are available only for x86_64, so if you use ARM e.g. Mac M1/M2, you must compile the app manually.  
 
-**Warning**  
-Prebuilt binaries are available only for x86_64, so if you use ARM machine like e.g. Mac M1, you need to compile manually app.  
+**Compilation is simple, and takes less than 10 minutes on an M1/M2 mac.**
```
alexisLefebvre, line 48: Out of curiosity, do you know why it's so long despite using these powerful M1/M2 CPU? Does it involve non-native code? Installing dependencies and compiling czkawka with my 8 cores / 16 threads CPU takes 2 minutes on Ubuntu 22.04, I used Docker so it's a little bit slower than direct compilation.

aioue: Downloading libraries

instructions/Installation.md
```diff
@@ -42,16 +42,21 @@ At the end execute it:
...
 
+User [@bauchdj](https://github.com/bauchdj) wrote [working rust complication instuctions](https://github.com/qarmin/czkawka/issues/689#issuecomment-1310549801)
```
qarmin: Steps should be pasted directly into this place to be able to update this steps if needed.

aioue: @qarmin steps pasted 👍

qarmin:
Thanks!

Issue #859, opened 2022-10-29
Upgrade Snap version to match latest release


santiagofn:
Currently [the version in Snapcraft](https://snapcraft.io/czkawka) is 4.1.0, while the app latest release is 5.0.2.

Snapcraft page should be updated.

orthorhombic:
It looks like [snapcraft.yaml](https://github.com/qarmin/czkawka/blob/27410bcd45b2b00dc88640cef97cb4bb4dc36920/snap/snapcraft.yaml) has been updated with the newest release (5.1.0)

@qarmin - If you're expecting this to automatically update, can you check your snapcraft account? Something might need your intervention/upload. Also, thanks for the great software!

qarmin:
Automatic updates are enabled, but there is still problem with compiling app
```
:: error: failed to compile `czkawka_gui v5.1.0 (/build/czkawka/parts/czkawka/build/czkawka_gui)`, intermediate artifacts can be found at `/build/czkawka/parts/czkawka/build/target`
::
:: Caused by:
::   package `gtk4-sys v0.6.0` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.61.0
Failed to run the build script for part 'czkawka'.
```
I cannot find, how to use in snap specific(latest) rust version, so until this one issue will be fixed, there will not be snap package(this is probably latest remaining issue)

santiagofn:
@qarmin I pocked around a little (... well, more than a little) and was able to fix it! 🎉  I created a PR with the solution.

PR #965, opened 2023-04-15
Fix Snap build

santiagofn:
Rustc & cargo are now installed using the script provided by rust-lang.org. Their versions were bumped to 1.68.2. The (outdated) versions bundled to core22 are not used anymore.

Solution copied from https://forum.snapcraft.io/t/correct-snapcraft-syntax-for-rust-based-snap-on-core22/30947/5

Fixes https://github.com/qarmin/czkawka/issues/859


qarmin:
Let's see if this will work(snaprcraft not work locally so if this works, this will be visible in edge channel)

qarmin:
Thanks!

PR #971, opened 2023-05-06
Add additional package for Mac installation

Liru:
cairo-rs fails building if `pkg-config` isn't installed on a Mac. It's fairly visible and easy to fix, but may as well add it to the instructions.


qarmin:
Thanks!

PR #1017, opened 2023-06-30
Fixed the number of similar images/videos per group in saved results.

jlesage:
I guess that the intend of the code was to print is the number of images/videos within the current group.


qarmin:
Thanks!

Issue #1045, opened 2023-08-18
Clear cache button typo


nfgallimore:
In settings there is a small typo:

"Remove outdated results from image cache" should be changed to "videos" cache

![image](https://github.com/qarmin/czkawka/assets/4875906/4aa28bb0-3e99-41fa-81fc-ceed355fe4ec)

https://github.com/qarmin/czkawka/blob/a77cb27f3b6d5818925fd6610631a0b6415ac78d/czkawka_gui/i18n/en/czkawka_gui.ftl#L385

PR #1047, opened 2023-08-18
Generalize clear cache button text 

nfgallimore:
Same constant is used for image and video clear cache button, updating to generalize text string. Fix for https://github.com/qarmin/czkawka/issues/1045


nfgallimore:
https://github.com/qarmin/czkawka/issues/1045

PR #1056, opened 2023-09-06
Improve Chinese translation

riokaa:
The original file was translated using machine translation, which made it difficult to understand. Therefore, I improved part of the translation.


qarmin:
Thanks, but translations should be changed in https://crowdin.com/project/czkawka (it is pinned in issue tracker), because updates are not bidirectional

PR #1106, opened 2023-10-23
Enable drag/drop for folder selection

akx:
This PR enables being able to drop files and directories from e.g. Finder into the include/exclude boxes.

I only tested this on Mac so far (dragging from the Finder nav bar or the Finder file list).

Both files and directories can be dragged into the views; when dragging files, the parent directory is added.


qarmin:
Tested and this works fine, thanks

PR #1109, opened 2023-10-24
Exclude container stuff by default on Linux

rugk:
STR:
1. Install/use podman or even better, Fedora Silverblue, where this is built-in.

You will see a lot of duplicates…

image

This should fix it. Note I used `*/home` to allow `/var/home` where Fedora Silverblue mounts it's homedir (it is symbolic-linked to `/home`, but I don't know how this app handles that.)

### DIFF
--- a/czkawka_core/src/common_items.rs
+++ b/czkawka_core/src/common_items.rs
@@ -3,7 +3,7 @@ use std::path::Path;
 use crate::common::Common;
 
 #[cfg(target_family = "unix")]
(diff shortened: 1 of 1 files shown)


alexislefebvre:
What path are you scanning? Does it work if you scan only `/home`?

rugk:
I scanned `/home/rugk` and no it does not work when I scan `/home` instead. Also does not work if I change it to `/home/*/.local/share/containers/*` to avoid the symblinks.

grafik

alexislefebvre:
Fedora Silverblue creates some symbolic links from `/var/home/` to `/home/`? It may help if Czkawka had a setting to now follow the links.

PR #1111, opened 2023-10-24
Improve bug reporting template

rugk:
I tried some adjustments to the bug report template to include more details and structure the information in a better way.

Let me know what you think. I personally e.g. also like headings (`##`) more than this `**let's make this bold as a heading:**` thing, but I used/tried to keep it similar to your style.


qarmin:
Thanks!

PR #1123, opened 2023-11-06
Typo in german translation

schmik:


alexislefebvre:
Translations should be done through Crowdin, see https://github.com/qarmin/czkawka/blob/master/instructions/Translations.md

PR #1124, opened 2023-11-06
Add basic zh_TW Traditional Chinese i18n locale

PeterDaveHello:
This is an early version but should be good enough to work, some minor adjustments may follow in the future :tada:


alexislefebvre:
Translations should be done through Crowdin, see https://github.com/qarmin/czkawka/blob/master/instructions/Translations.md

PeterDaveHello:
Oh, okay, let me upload to there! Thanks!

alexislefebvre:
My bad, actually I don't know if Crowdin can accept a new language. Let's wait for an update from @qarmin

PeterDaveHello:
Sure, it'll be also very convenient if the first add can be done here on GitHub :star_struck:

qarmin:
With czkawka I already hit 60k word limit, so I haven't had much opportunity to add new languages, but yesterday I managed to get the status of an open source project so the limit should no longer cause problems.

I imported the translations and it looks like everything is working there - https://crowdin.com/project/czkawka/zh-TW

The documentation to adding a new language to app is pretty much just for me(I need to change/hide it later), and all modifications should be done mainly in Crowdin.

The language should be available in the new version of the app, as long as I don't forget to do it for myself, as I forgot with version 6.1.

Thanks for your efforts

Issue #4, opened 2020-09-22
Translation support


qarmin:
It would be great to see translation support in czkawka (both orbtk and gtk frontend)

err931:
@qarmin 
[How to bind a text domain to a local folder for gettext under GTK3](https://stackoverflow.com/questions/10094335/how-to-bind-a-text-domain-to-a-local-folder-for-gettext-under-gtk3)

The above post is about supporting multiple languages in Python, but you can use [gettext-rs](https://crates.io/crates/gettext-rs), to do the same thing.
There is also a struct called [TextDomain](https://docs.rs/gettext-rs/0.6.0/gettextrs/struct.TextDomain.html).

I hope this will help.

alexislefebvre:
Could https://crowdin.com/ help people to translate the strings? It will require some time to set it up but users will be able to translate the project without any tool, just by using the Crowdin website.

qarmin:
For now I don't see a problem with current file based translations, since only 3 lanugages at this moment are supported(2 of them I maintain)
But with more languages this will be a mess.

I checked crowdin, but I see that they have limit 60000 words on free plan.
This may be a problem if are counted all translations(for now single one translation takes 2500/3000 words(checked by `wc` tool) but in future this may increase a lot).

I'm not sure if Fluent is fully supported(but use same key-value approach like gitea, so probably is)

LAfricain:
I'm working on French.

qarmin:
French language was already added - https://crowdin.com/project/czkawka/fr, but this is computer translation, so needs to be fixed because in some places probably looks like "I be hungry"

igor-cali:
@qarmin Any plan to release the update with available translations? I am specifically interested to the Ubuntu package from PPA. Thanks in advance for your insights.

qarmin:
I plan to release 4.0.0 version at beginning of February.
All features which I wanted to include in this release, are already merged.
Only some bug fixes are required, but probably a.
3.3.1 was released 1,5 months ago, so I don't want to publish new version too often.

For now there are 3 methods to get the latest version:
- Binaries from actions
- Arch git binaries
- Snap package(similar video not works)

alexislefebvre:
> 3.3.1 was released 1,5 months ago, so I don't want to publish new version too often.

Maybe you could release an alpha version (and mark it as “pre-release” in GitHub), it would help to [Release early, release often](https://en.wikipedia.org/wiki/Release_early,_release_often).

[...]

PR #1154, opened 2023-12-10
Revise some translations

yimengqinyao:
Revise some translations


qarmin:
Thanks,

but translation should be updated on crowdin page accoring to instruction - https://github.com/qarmin/czkawka/blob/master/instructions/Translations.md#translations / pinned issue - https://github.com/qarmin/czkawka/issues/4.

Two way translation synchronization is hard, so I only support translations on this site

PR #1209, opened 2024-02-14
Use sets in the tests for expected and actual files to make those stable

blob79:
Rather use sets in the tests for expected and actual files to make those stable.


blob79:
Ah this was a bit to quick. Closing.

qarmin:
Thanks!

PR #1250, opened 2024-03-21
Improve help messages

waldyrious:
Various small changes to the help messages to make them more readable and consistent, correct some typos, expand missing information, etc.)

### DIFF
--- a/czkawka_cli/src/commands.rs
+++ b/czkawka_cli/src/commands.rs
@@ -94,7 +94,7 @@ pub struct DuplicatesArgs {
         value_parser = parse_minimal_file_size,
         default_value = "8192",
         help = "Minimum size in bytes",
-        long_help = "Minimum size of checked files in bytes, assigning bigger value may speed up searching"
+        long_help = "Minimum size of checked files in bytes; assigning bigger values may speed up searching"
     )]
     pub minimal_file_size: u64,
     #[clap(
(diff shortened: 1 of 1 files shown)


czkawka_cli/src/commands.rs
```diff
@@ -197,16 +197,16 @@ pub struct SimilarImagesArgs {
...
-        long_help = "Maximum size of checked files in bytes, assigning lower value may speed up searching"
+        long_help = "Maximum size of checked files in bytes; assigning lower values may speed up searching"
     )]
     pub maximal_file_size: u64,
     #[clap(
         short,
         long,
         default_value = "High",
         value_parser = parse_similar_images_similarity,
-        help = "Similairty level (Minimal, VerySmall, Small, Medium, High, VeryHigh, Original)",
```
skywinder, line 208: looks   Original is missed?

waldyrious: `Original` is currently not supported as a value that can be passed:

https://github.com/qarmin/czkawka/blob/2a32a52aa882f6ff52ba4d3e24a666dc2a86cf9b/czkawka_cli/src/commands.rs#L575-L585

I was actually planning to propose adding it as an option (perhaps renaming it to "Maximal" to make it more accurate) in a separate PR :)

waldyrious:
Hey @qarmin — could you share any notes about the reason for closing?

PR #1252, opened 2024-03-24
macOS app bundle creation

skywinder:
**Application Bundle Creation**: The application bundle structure creation and binary copying processes have been streamlined for efficiency.

Update documentation.

1. Install `librsvg` with Homebrew: `brew install librsvg`
2. Run the script: `./misc/create_app_bundle.sh`
3. The `.app` bundle will be in `./target/release`
4. Move the bundle to Applications: `mv ./target/release/Czkawka.app /Applications`

CleanShot 2024-03-24 at 20 16 21@2x

Enjoy Czkawka.app


qarmin:
Can you add this to mac CI?

skywinder:
@qarmin I miss your comment, sad that it's just closed. It could be useful for many users for sure. If you need Mac CI, I can try to do this.

qarmin:
That would be great

I could theoretically add macOS package generation to CI, but without access to a Mac for testing, I prefer not to risk shipping something that might be broken.

https://github.com/qarmin/czkawka/blob/master/.github/workflows/mac.yml

The script needs to be adjusted slightly to handle both czkawka_gui and krokiet, including their different logos (there are also multiple variants of these apps). It should also handle external dependencies such as libraw, gtk, etc. - this is currently the biggest pain point for macOS users, which try to run app

Issue #1118, opened 2023-10-28
Delete All Expect Largest? In Cli


someonenameguy:
This is possible to do in the Cli?
Would be useful for Images.

PR #1255, opened 2024-03-25
Add new delete methods based on size

slgeay:
This adds 4 new delete methods:

- AEB: all except biggest
- AES: all except smallest
- OB: one biggest
- OS: one smallest

From a quick look at the issues it should implement #1118


qarmin:
Thanks!

PR #1261, opened 2024-03-30
doc: add link to Dockerized version

alexislefebvre:
This is great to run the GUI version, it works on a headless server too.


alexislefebvre:
The link is already present: https://github.com/qarmin/czkawka/blob/c6b1eaeeb402a4e94263ccc2ab64b11277537c3b/instructions/Installation.md?plain=1#L171

PR #1275, opened 2024-05-02
initial work for getting a 7.0.0 build to flathub

Renner0E:
Hey all,

I made some changes to maybe(?) successfully push the 7.0.0 build to flathub.
Right now it fails on validation because flathub changed their [metadata guidelines](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/) sometime in the past months.

[Related PR to update the flatpak manifest](https://github.com/flathub/com.github.qarmin.czkawka/pull/26) and https://buildbot.flathub.org/#/builders/25/builds/14554

developer_name is deprecated but it cannot be removed because some frontends do not support the new developer tag yet see: 
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer
https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#name-summary-and-developer-name

Verify the metainfo.xml offline

git clone https://github.com/qarmin/czkawka && cd czkawka

`appstreamcli validate --explain data/com.github.qarmin.czkawka.metainfo.xml`
or 
`flatpak run --command=flatpak-builder-lint org.flatpak.Builder appstream data/com.github.qarmin.czkawka.metainfo.xml`

ERROR before any changes:
```
I: com.github.qarmin.czkawka:5: summary-has-dot-suffix
Multi functional app to find duplicates, empty folders, similar images, broken files etc.
The component summary should not end with a dot (`.`).

E: com.github.qarmin.czkawka:17: screenshot-image-source-duplicated
There can only be one `source` image per screenshot and language.

E: com.github.qarmin.czkawka:18: screenshot-image-source-duplicated
There can only be one `source` image per screenshot and language.

I: com.github.qarmin.czkawka:25: developer-name-tag-deprecated
The toplevel `developer_name` element is deprecated. Please use the `name` element in a
`developer` block instead.

I: com.github.qarmin.czkawka:~: developer-info-missing
This component contains no `developer` element with information about its author.

✘ Validation failed: errors: 2, infos: 3, pedantic: 1

```

Maybe also consider changing the app id from com.github.qarmin.czkawka to io.github.qarmin.czkawka. See https://docs.flatpak.org/en/latest/conventions.html


qarmin:
Thanks

PR #1288, opened 2024-05-25
Print error to stdout if reading a heic image fails

andreasgerstmayr:
I tried to de-duplicate similar heic images, but unfortunately they were skipped without any indication of an error. Printing the error to the console helps users resolving the error.


qarmin:
This is not proper fix, because will break CLI by printing random data when terminal draws progress bar.
This info should go to Messages struct, will check later into this

qarmin:
Changed this in https://github.com/qarmin/czkawka/pull/1355

PR #1325, opened 2024-08-16
Fix build regression in rust 1.80 due to time crate.

iynaix:
Rust 1.80 contains a [build regression](https://github.com/rust-lang/rust/issues/127343) that causes the `time` crate to fail to build. This is a simple update to use a newer version of the time crate (originally patched for NixOS/nixpkgs#335000)


qarmin:
Superseded by #1354

PR #1350, opened 2024-09-21
Use static runtime, make zsync file and use zstd compression

### LINKED ISSUE #1347: libfuse2 is EOL for appimages
**Feature Description**
libfuse2 is EOL for appimage version of this app. Please migrate to libfuse3. Thank you.

Samueru-sama:
These changes get rid of the libfuse2 dependency and also result in a slightly smaller AppImage that also starts up faster.

Fixes #1347 

Adding the zsync files allows for delta updates with [AppImageUpdate](https://github.com/AppImageCommunity/AppImageUpdate)

Note that this will only work for the "alternative" AppImage, if you want the other appimage to get updates I'm not sure if it is possible to tell pkg2appimage to not create the appimage, it likely means extracting the appimage and rerunning it with appimagetool, it also means that the zsync info in both appimages has to change to prevent conflicts during updates, as right now appimageupdate will look for a `*.AppImage.zsync` when updating the AppImage, which as long as there is only one zsync file it isn't a problem.

### DIFF
--- a/.github/workflows/linux_gui.yml
+++ b/.github/workflows/linux_gui.yml
@@ -135,18 +135,29 @@ jobs:
           pwd
           wget -c "https://raw.githubusercontent.com/linuxdeploy/linuxdeploy-plugin-gtk/master/linuxdeploy-plugin-gtk.sh"
           wget -c "https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage"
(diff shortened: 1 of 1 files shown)


Samueru-sama:
I noticed the ci failed because I had a typo when calling `appimagetool-x86_64.AppImage` as just `appimagetool` 🫠

I'm very sorry.

PR #1356, opened 2024-09-27
New version video similarity library

qarmin:
This version allows to compare similarity of video files, that are shorter than 30 seconds

Not mergable yet, because currently app depends on library from commit https://github.com/Farmadupe/vid_dup_finder_lib/commit/4015aed8efc83bf9f6643158bc714f0767d6de8a(czkawka only can use version released on crates.io)

Second problem are random panics, library needs to be tested with fuzzer - recent panic I found with integer overflow(trying to catch it, but it is also possible, that I used wrong version and all panics are already fixed).

Also current version, not allows to check if ffmpeg is installed

Image-rs still is used in version 0.24 - this causes version duplication and czkawka user needs to compile both 0.24 and 0.25 versions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,7 +9,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "79faae4620f45232f599d9bc7b290f88247a0834162c4495ab2f02d60004adfb"
 dependencies = [
  "ab_glyph_rasterizer",
- "owned_ttf_parser",
+ "owned_ttf_parser 0.24.0",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


Farmadupe:
Currently working on getting a draft of the next version of vid_dup_finder_lib pushed onto github and/or crates.io.

This version should fix the "too short" issue.

Farmadupe:
* Updated version of library pushed to main branch in github [github](https://github.com/Farmadupe/vid_dup_finder_lib/tree/main/vid_dup_finder_lib)
* Published vid_dup_finder_lib  v0.2.0 https://crates.io/crates/vid_dup_finder_lib to crates.io
* API is slightly different. Instead of `VideoHash::new(some_path)`, hashes are now build through a factory `VideoHashBuilder::default().hash(some_path)`
    * ([Latest docs](https://docs.rs/vid_dup_finder_lib/0.2.0/vid_dup_finder_lib/) have a doctestable example of how to use new API)

If you are able to integrate with Czkawka, hopefully I should have some time during holidays to assist with overflow bug.

qarmin:
Thanks, I was able to integrate vid_dup_finder library with fuzzer and I found 2 problems that already reported to repo. 

Library will be updated in https://github.com/qarmin/czkawka/pull/1425

PR #1357, opened 2024-10-01
increase hashing buffer size

i512:
Hi guys, thanks for your project.
Full hashing was taking a lot of time on my HDD, and task manager reported read speed was kinda low, ~25MB/s so I suspected the small buffer might cause it.

Tested with 16MB buffer read speed increased to ~150MB/s which is what I would expect.
20 gb of duplicates is scanned in 3 minutes vs 13 on current release (without cache).

Interestingly if I run the same test on an NVMe then there's almost no difference: 17s vs 18s, ~2.4GB/s read speed. 
I guess windows is reading small blocks from the drive without doing a lot of readahead. Increasing the buffer is kind of a simple fix.

### DIFF
--- a/czkawka_core/src/duplicate.rs
+++ b/czkawka_core/src/duplicate.rs
@@ -777,7 +777,7 @@ impl DuplicateFinder {
             .map(|(size, vec_file_entry)| {
                 let mut hashmap_with_hash: BTreeMap<String, Vec<DuplicateEntry>> = Default::default();
                 let mut errors: Vec<String> = Vec::new();
-                let mut buffer = [0u8; 1024 * 16];
+                let mut buffer = vec![0u8; 1024 * 1024 * 16];
 
                 atomic_counter.fetch_add(vec_file_entry.len(), Ordering::Relaxed);
                 for mut file_entry in vec_file_entry {


qarmin:
This is a little strange. Using stack should be faster than allocating each time buffer on heap(but probably hdd is bigger bottleneck here).
Previous version used here 128KB array, but there was a lot of stack overflow so I decreased value to fix problems.

This may improve performance for bigger files, but I doubt that allocating 16MB of memory even for small 5KB files will not decrease performation in such situation.

To verify that a benchmark for function hash_calculation should be created.

I think that mixed solution, based on file size, should works better, but without clear benchmark with different parameters, we cannot be sure.

i512:
Sorry, I'm new to rust, didn't realize buffer would be allocated each time. That's not good.

> I think that mixed solution, based on file size, should works better, but without clear benchmark with different parameters, we cannot be sure.

This can work, but it will require tuning, yes. 

I think the usual approach is to reuse the buffers between jobs. I can't figure out how to do this yet though. Tried creating buffers in `map_init`, but I did not work as I expected. So I'm probably going to close this for now.

Reducing the number of threads to 1 (in current release with 16kb buffer) actually improves the read speed with big files in my case. Oh I see that this was already mentioned in https://github.com/qarmin/czkawka/issues/835 nice.

PR #1400, opened 2024-11-21
fix: czkawka-gui window icon under Wayland

Integral-Tech:
Currently, when running czkawka-gui under Wayland, the window icon will fallback to default Wayland icon. Set application ID to fix this problem.


Integral-Tech:
@qarmin

PR #1444, opened 2025-02-05
print currently decoding file to debug

Quackdoc:
Helpful for finding issues when czkawka fails to decode a file or hash it for whatever reason


qarmin:
I changed this to trace, to not pollute too much debug logs

PR #1478, opened 2025-03-17
Converts unnecessary println! into info/error etc.

### LINKED ISSUE #1029: Reshot free license isnt "free"
https://www.reshot.com/license/

The license has a relevant caveat that was left out in the License file.
> There are some important limits to these rights, described in our [User Terms](https://www.reshot.com/terms/).

> 2. Your use of our Items is subject to our Reshot License. There are a couple of important limits to your license rights: you must not (a) use an Item to build a service or product that is similar or competitive with Reshot (or try to); (b) aggregate or collate an Item(s) and make available on a stock or inventory basis or in a library or other collection for use or download; (c) use an Item in the context of adult content (such as nudity or sexual services), supply of weapons or gambling services, exploitation of children, hate speech, defamatory content, misleading or deceptive conduct, or any illegal activity; (d) sell physical or digital copies of Items without first altering them by applying skill and effort, and incorporating other elements (merely printing an Item on an object such as a mug or shirt is not sufficient alteration); or (e) register or use an Item as a trademark or otherwise claim ownership or registered protection over an Item. The provisions of the Reshot License continue to apply to all use of the Item, even when you sub-license or transfer it.

This effects the freedoms given by the MIT license and would block a transition to [GPL-3](https://github.com/qarmin/czkawka/issues/347) 

(I will note that im not a lawyer or qualified in any way)

https://freedomdefined.org/Definition

qarmin:
- Converts println into log functions
- Speedup a little cache loading/saving
- Allow to select more than one item in krokiet with support for mouse and keyboard
- Replaces icons in gui to free ones

Closes #1029


czkawka_gui/LICENSE
```diff
@@ -32,6 +33,16 @@ CC BY 4.0 License
...
 Copyright (c) 2020 [jannuary](https://github.com/jannuary)
 
+and
+
+Icons
+czkawka_gui/icons/*.svg
+krokiet/icons/*.svg
+krokiet/icons/*.png
+
+Copyright (c) 2020-2025 Rafał Mikrut
```
tenzap, line 43: Great! So what is the license for the icons then? MIT ? CC BY 4.0 License

qarmin: The icons are licensed under the CC BY 4.0 License (it is like MIT, but for images)

I chose this license because it's specifically designed for creative works like icons, images, and other visual assets — unlike the MIT license, which is mainly intended for software/code(at least I think so).

tenzap: ok, thanks for clarifying.

tenzap: What is actually the license of `icons/icon_about.png` ? It looks derived from `data/icons/com.github.qarmin.czkawka.svg` and is not covered by `czkawka_gui/icons/*.svg` of your LICENSE file (and previously the license file didn't restrict to *.svg, but all files in `icons`). Therefore it is difficult to know if it is MIT or CC-BY-4.0 (which doesn't require derived work from having the same license).

qarmin: All icons now, are licensed CC-BY-4.0 

https://github.com/qarmin/czkawka/blob/8961c0a8ce16978f014cc4f8847d883b505ed711/czkawka_gui/LICENSE#L27-L46

tenzap: If I'm not mistaken, this list doesnt cover icons/icon_about.png which is a rename of a file originally in snap/ (at least in v 4.0)
The filter in license file mentions icons/*.svg only

qarmin: Well, yes, I manually created this one PNG from an SVG, because GTK had some issues displaying it. However, it's licensed the same way as the SVG files. I'll add a note about that in one of the next commits

tenzap: Ok.
So just to be sure and in order to reference your reply for the Debian package of 8.0, since the first time this file was committed, you intended it to be under cc-by-4.0, right?

qarmin: Yes, just like any other icon now

PR #1480, opened 2025-03-18
Allow to disable similar videos, which are broken on 32 bit os

qarmin:
This adds a "Similar Videos" tool as an optional feature, enabled by default.

I don't think this PR should be merged into the main branch because:

- The "Similar Videos" tab cannot be clearly disabled in gtk
- It currently does not work with Krokiet
- Disabling it does not provide significant benefits (e.g., a lot of faster compilation speed)


qarmin:
Latest version restore support for 32 bit

Issue #1631, opened 2025-09-16
[macos+linux] tiny image previews


fenbyte:
**Bug Description**
image previews on macos (and supposedly linux) are unreasonably tiny

**Steps to reproduce:**

<img width="1868" height="1204" alt="Image" src="https://github.com/user-attachments/assets/969b8db1-1316-4738-b4dd-5a5c7f3fccd9" />

**System**

- Czkawka/Krokiet version:  Czkawka gtk version: 10.0.0, release mode, rust 1.86.0 (2025-03-31), os Mac OS 26.0.0 (aarch64 64-bit), 8 cpu/threads, features(2): [heif, libavif], app cpu version: Baseline, os cpu version: Baseline
- OS version:  macOS 26 Tahoe arm64
- Installation method:  nightly with heif+avif

rv-arteaga:
The same error occurs in Arch Linux. Regards.

Soblow:
Same issue here (Linux)

- Czkawka gtk version: 10.0.0, release mode, rust 1.85.0 (2025-02-17), os Arch Linux Rolling Release (x86_64 64-bit), 16 cpu/threads, features(2): [heif, libraw], app cpu version: x86-64-v1 (SSE2), os cpu version: x86-64-v4 (AVX-512)
- Arch Linux (rolling release) Linux 6.16.8-arch3-1
- Installation method: Arch Linux **AUR** (also tested the **standalone binary** and a manual **build from sources**, same issue)
- System GTK version: 4.20.1-1

The Appimage of 9.0 works as intended and doesn't have this issue.
Noting the fact that I used 10.0 for some time before this issue appeared, so it most likely comes from a dependency that updated and broke something

Only one error is logged in the console, but was already here before:
```
(com.github.qarmin.czkawka:171924): Gtk-CRITICAL **: 16:17:22.066: gtk_css_node_insert_after: assertion 'previous_sibling == NULL || previous_sibling->parent == parent' failed
```

Emporea:
same issue here

my guess is https://github.com/qarmin/czkawka/pull/1590

Soblow:
> my guess is [#1590](https://github.com/qarmin/czkawka/pull/1590)

Reverting commit 5dc925737530dc6bb6c106c336c6d49e6066f0e6 doesn't fix the issue on my end (recompiled from sources)

Emporea:
I didnt even test it before guessing, but you are right. it didnt change it.

ghost:
[...]

PR #1644, opened 2025-10-04
temp fix for #1631

0000xFFFF:
temp fix for #1631


qarmin:
Thanks, but I applied the fix suggested by the GTK developers — changing GtkImage to GtkPicture

Issue #1652, opened 2025-10-15
Similar images results are non-deterministic


RReverser:
**Bug Description**

**Steps to reproduce:**

Searching for duplicates over and over keeps grouping images differently, which can be pretty confusing. In this screencast you can see how reruns result in IMG_4329.jpg being grouped not only in different order, but with different images and even in groups of different size each time.

I noticed this first when I was cleaning images, running scan again, got a lot more duplicate results, cleaned those up, clicked scan again, got more duplicate results again, and so on until it actually runs out.

Since this uses image hashing, I'd expect image groups to be always the same and all duplicates to be returned in one go.

https://github.com/user-attachments/assets/fee9acaf-d6eb-45b6-8c11-78177dd0127d

**System**

2025-10-14 16:01:13.653 [INFO] [czkawka_core::common::logger] (1) logger.rs:142 - Krokiet version: 10.0.0, release mode, rust 1.85.0 (2025-02-17), os Windows 10.0.26100 (x86_64 64-bit), 22 cpu/threads, features(0): [], app cpu version: x86-64-v1 (SSE2), os cpu version: x86-64-v3 (AVX2)
2025-10-14 16:01:13.653 [INFO] [czkawka_core::common::config_cache_path] (1) config_cache_path.rs:155 - Config folder set to "C:\Users\me\AppData\Roaming\Qarmin\Krokiet\config" and cache folder set to "C:\Users\me\AppData\Local\Qarmin\Czkawka\cache"
2025-10-14 16:01:13.653 [INFO] [krokiet] (1) main.rs:155 - Krokiet features(2): [winit_femtovg, winit_software]
2025-10-14 16:01:13.973 [INFO] [krokiet::connect_translation] (1) connect_translation.rs:145 - System language: en
2025-10-14 16:01:13.973 [INFO] [krokiet::connect_translation] (1) connect_translation.rs:148 - Language: en

PR #1654, opened 2025-10-15
Make comparison results deterministic in the UI

RReverser:
Demo (left is release version, right is the one built from this branch):

https://github.com/user-attachments/assets/e7c8913a-aa21-4dd4-8394-97e1662535b4

Fixes https://github.com/qarmin/czkawka/issues/1652.

The issue was that the code uses the standard HashMap in a lot of places, which infamously has randomised iteration order as a protection from DoS attacks on servers.

These concerns don't apply to local apps, and the randomisation leaks into comparisons and the UI.

To fix it, I've switched all the usages of HashMap and HashSet in czkawka_core to the indexmap crate, which is intended exactly for this kind of scenarios - when you want the performance of the HashMap, but care about deterministic iteration.

### DIFF
--- a/czkawka_core/src/localizer_core.rs
+++ b/czkawka_core/src/localizer_core.rs
@@ -1,6 +1,7 @@
+use std::collections::HashMap;
+
 use i18n_embed::fluent::{FluentLanguageLoader, fluent_language_loader};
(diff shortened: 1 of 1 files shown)


RReverser:
The remaining CI issues seem unrelated to my PR.

PR #1703, opened 2025-12-08
Refine Chinese translations in krokiet.ftl

OlingCat:


qarmin:
Translations should only be modified on Crowdin, because introducing two-way synchronization would require conflict resolution between GitHub and Crowdin. This would significantly complicate the workflow and could lead to merge issues and broken translations.

For this reason, please submit all translation changes directly on Crowdin.

https://github.com/qarmin/czkawka/blob/6296b06f3a6798f07f78cd771ed95d75639f7bbb/instructions/Translations.md?plain=1#L9-L13

PR #1738, opened 2026-01-10
Add shortest/longest path selection modes for czkawka and krokiet

JoBoCl:
Note: these changes are as consistent as possible with the existing code, rather than attempting to refactor the existing methods - I'd like to submit these refactors as follow-up PRs.

### DIFF
--- a/krokiet/src/connect_select.rs
+++ b/krokiet/src/connect_select.rs
@@ -116,9 +116,7 @@ fn extract_comparable_field(model: &MainListModel, property: Property, active_ta
             connect_i32_into_u64(high, low)
         }
         Property::PathLength => val_strs.nth(active_tab.get_str_path_idx()).expect("can find file path property").len() as u64,
-        Property::Resolution => {
-            val_ints.nth(active_tab.get_int_pixel_count_idx()).expect("can find pixel count proerty") as u64
-        }
+        Property::Resolution => val_ints.nth(active_tab.get_int_pixel_count_idx()).expect("can find pixel count proerty") as u64,
     }
 }
 


JoBoCl:
The refactor mentioned to reconcile the krokiet selection functions is available [here](https://github.com/JoBoCl/czkawka/commit/b0202661a0d806a2db910279556032aa5078c3a7).

qarmin:
This refactoring can be put here - I’ll be testing all the changes manually anyway, so it will be easier for me to test it once instead of twice.

JoBoCl:
Done - I've added the refactoring and fixed the clippy warnings.

I've not attempted to refactor the czkawka implementation - it seems somewhat more brittle, and I'm assuming that there's more emphasis on krokiet moving forwards. (I'd also add that developing for krokiet was much nicer)

krokiet/src/connect_select.rs
```diff
@@ -87,81 +101,35 @@ fn set_select_buttons(app: &MainWindow) {
...
-                }
-            }
-            if !old_data[min_item_idx].checked {
-                checked_items += 1;
-            }
-            old_data[min_item_idx].checked = true;
-        }
+fn extract_comparable_field(model: &MainListModel, property: Property, active_tab: ActiveTab) -> u64 {
+    let val_ints = model.val_int.iter().collect::<Vec<_>>();
+    let val_strs = model.val_str.iter().collect::<Vec<_>>();
```
qarmin: Collecting results into vectors only to discard half of them shortly afterwards (since only one of val_int or val_str can be used at a time), and then effectively using only part of the available values, is not very efficient. With a larger number of records, this results in noticeable slowdowns and is visible in perf stats.

Previously, results were also collected into a vector, but the difference was that they were collected into a single one.

The fastest fix would likely be to eliminate collecting elements into vectors altogether and instead use an iterator with nth, for example:
```
val_ints.iter().nth(active_tab.get_str_path_idx()).len() as u64
```

image

JoBoCl: Added in latest commit, thanks for pointing that out.

krokiet/src/connect_select.rs
```diff
@@ -87,81 +101,50 @@ fn set_select_buttons(app: &MainWindow) {
...
+        Property::Date => {
+            let high = val_ints.nth(active_tab.get_int_modification_date_idx()).expect("can find file last modified property");
+            let low = val_ints.next().expect("can find file last modified property");
+            connect_i32_into_u64(high, low)
+        }
+        Property::PathLength => val_strs.nth(active_tab.get_str_path_idx()).expect("can find file path property").len() as u64,
+        Property::Resolution => {
+            let h_idx = active_tab.get_int_height_idx();
+            let w_idx = active_tab.get_int_width_idx();
+            assert_ne! {h_idx, w_idx, "file height and width must be stored in separate properties"};
```
qarmin: After spending some time trying to understand what is going on here, I realized that this is an optimization where only a single iterator is created instead of two. In general, this is an acceptable approach in many cases, but in this form it is extremely hard to read.

In this case, the simplest solution would be to add a new function to active_tab, `get_int_pixel_count_idx()`, which would return the index of `IntDataSimilarImages::PixelsPart1`. The logic would then be the same as above in Property::Date.
```
            let high = val_ints.nth(active_tab.get_int_pixel_count_idx()).expect("can find file last pixel count property");
            let low = val_ints.next().expect("can find file last pixel count property");
            connect_i32_into_u64(high, low)
```

JoBoCl: Done

PR #1763, opened 2026-01-30
Fix builds for Android and iOS targets

ClementTsang:
Been trying to get this working with `flutter_rust_bridge` for an Android app I was playing around with, but right now I'm hitting an error like this:

```bash
SEVERE: STDERR:
SEVERE: Compiling trash v5.2.5
SEVERE: error[E0433]: failed to resolve: use of unresolved module or unlinked crate `platform`
SEVERE:   --> /home/me/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/trash-5.2.5/src/lib.rs:67:35
SEVERE:    |
SEVERE: 67 |         Self { platform_specific: platform::PlatformTrashContext::new() }
SEVERE:    |                                   ^^^^^^^^ use of unresolved module or unlinked crate `platform`
SEVERE:    |
SEVERE:    = help: if you wanted to use a crate named `platform`, use `cargo add platform` to add it to your `Cargo.toml`
```

It looks like it's because the `trash` crate [does not support Android or iOS](https://github.com/Byron/trash-rs/blob/master/src/lib.rs#L44-L55):

```rust
#[cfg(target_os = "windows")]
#[path = "windows.rs"]
mod platform;

#[cfg(all(unix, not(target_os = "macos"), not(target_os = "ios"), not(target_os = "android")))]
#[path = "freedesktop.rs"]
mod platform;

#[cfg(target_os = "macos")]
pub mod macos;
#[cfg(target_os = "macos")]
use macos as platform;
```

Also, at least for Android, it seems like it's unsupported there because it doesn't have the semblance of a standardized system trash, for example (https://github.com/Byron/trash-rs/issues/10).

This PR just makes it so that the `trash` crate dependency is skipped for iOS and Android targets, and adds a wrapper around the `trash::delete` (just called it `trash_delete`) to fall back to system deletion for iOS and Android.

Using this branch in my Android app successfully compiles now so I'm assuming that this is enough to make builds work for at least Android.

### DIFF
--- a/czkawka_core/src/common/mod.rs
+++ b/czkawka_core/src/common/mod.rs
@@ -128,8 +128,8 @@ pub fn check_if_folder_contains_only_empty_folders<P: AsRef<Path>>(path: P) -> R
     Ok(())
 }
 
-/// A wrapper around `trash::delete`, falling back to `fs::remove_dir_all` for platforms that do not support
-/// the library (e.g. Android, iOS).
+/// A wrapper around `trash::delete`. Note that for platforms that do not have native trash support
+/// (Android, iOS), this function will always return an [`Error`].
 fn trash_delete<P: AsRef<Path>>(path: P) -> Result<(), String> {
     let path = path.as_ref();
 
@@ -140,13 +140,15 @@ fn trash_delete<P: AsRef<Path>>(path: P) -> Result<(), String> {
 
     #[cfg(any(target_os = "android", target_os = "ios"))]
     {
-        fs::remove_dir_all(path).map_err(|err| err.to_string())
+        Err("trash is not supported on this platform".to_string())
     }
 }
 
 /// Remove the folder if it only contains empty folders/is empty. If `remove_to_trash` is set, the folder
 /// will instead be sent to the system's recycle bin/trash equivalent rather than being deleted.
-/// Note that this will not do anything if the system does not have native trash support (e.g. iOS, Android).
+///
+/// Note: if used on Android or iOS platforms, ensure `remove_to_trash` is false, as trash is not supported
+/// and will always return an [`Error`].
 pub fn remove_folder_if_contains_only_empty_folders<P: AsRef<Path>>(path: P, remove_to_trash: bool) -> Result<(), String> {
     check_if_folder_contains_only_empty_folders(&path)?;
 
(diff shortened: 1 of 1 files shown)


czkawka_core/src/common/mod.rs
```diff
@@ -128,26 +128,44 @@ pub fn check_if_folder_contains_only_empty_folders<P: AsRef<Path>>(path: P) -> R
...
+    let path = path.as_ref();
+
+    #[cfg(not(any(target_os = "android", target_os = "ios")))]
+    {
+        trash::delete(path).map_err(|err| err.to_string())
+    }
+
+    #[cfg(any(target_os = "android", target_os = "ios"))]
+    {
+        fs::remove_dir_all(path).map_err(|err| err.to_string())
```
qarmin: This will fail on android/ios

```
remove_dir_all will fail if the path is not a directory.
```
https://doc.rust-lang.org/std/fs/fn.remove_dir_all.html

different wrappers should be added for file/folder

qarmin: Alternatively, the Android and iOS code paths may contain an unreachable!() macro. In addition, a warning can be added above the functio stating that trash support is not available on these platforms and that the caller is responsible for ensuring that the remove_to_trash option is not enabled there.

ClementTsang: Makes sense, will change.

ClementTsang: Hm, would it make sense to simply just return an `Error` if you call `trash_delete` on Android/iOS, rather than panicking with `unreachable`? I'll still add that yes, it'll unconditionally fail for those platforms.

qarmin: An error is also acceptable. It would behave the same as a panic, just in a less dramatic way. For me, a panic in this place would not be an issue, because very few people compile Krokiet or the CLI version for iOS or Android, so it is unlikely that a random user would accidentally hit this panic. In practice, only application developers would encounter it, which would force them to disable the remove_to_trash option more quickly, since it would be immediately obvious that it is unusable option for them.

I will try to add a simple CI check that verifies Android compilation on merge to the master branch, to prevent similar build issues in the future.

ClementTsang: Thanks!

ClementTsang: > CI check

Also let me know if you would like me to help with this.

PR #1767, opened 2026-01-31
Add GPU detection and video processing enhancements

Wadelz:
Introduce GPU detection, codec detection, and a decoder strategy for improved video processing. Implement a pre-scan module for GPU compatibility analysis and a temporal segmentation module for chunked video hashing. Include comprehensive unit tests for all new features and address code review feedback. Document the implementation plan for future improvements.


qarmin:
It looks like the entire PR is AI-generated. I have no issue with that as long as it works and has consistent styling, but unfortunately it does not.

Since this is not marked as a draft PR, I assume it is ready for review. In that case:

- It does not work - the code does not compile.
- It does not work (2) - some files were added to the `common` directory and a few parameters to the `Video` struct, but nothing beyond that. There is no end-to-end flow that I can build or test.
- I doubt that this will work fine - finding good video similarity algorithm is hard and most known approaches have serious trade-offs. With the current solution, I expect both performance and similarity quality to be poor. I can be wrong, but I cannot verify this, because the code is not functional.
- The entire logic should be extracted into an external crate  -  `czkawka_core` is intended to be glue code that connects dependencies behind a unified interface usable by other projects. For example, similar videos are using ~20 lines of code, from imported vid_dup_finder_lib, without needing to maintain the internals. With the current implementation, this responsibility to maintain this code would fall on me, which is not acceptable.
- No AMD support - this is a smaller issue, but I use AMD GPUs and they do support hardware decoding for common formats.

So in short, please extract this into an external crate and in this PR, include only the glue between czkawka and this library. I will test whether it works and if it works well, I will help improve both your library and its integration with the CLI and GUI.

qarmin:
Closing this PR, because it appears that you are at the beginning of your programming journey and this topic is too advanced at this stage, at least for now.
Feel free to open a new PR once all of the issues mentioned above have been addressed.

PR #1780, opened 2026-02-07
Use Slint 1.15.0 also in Krokiet/Cargo.toml (not just flatpak) + make tree_iter refs mutable where gtk4-rs api requires

thomas725:
this pull request is meant to replace my previous one: https://github.com/qarmin/czkawka/pull/1779
which failed to build for my NixOS package

### DIFF
--- a/krokiet/Cargo.toml
+++ b/krokiet/Cargo.toml
@@ -32,10 +32,15 @@ i18n-embed = { version = "0.16", features = ["fluent-system", "desktop-requester
 i18n-embed-fl = "0.10"
 rust-embed = { version = "8.5", features = ["debug-embed"] }
 
+
+# qarmin: To fix 32 bit CI, this dependency must be added:
+# Easier cross-compilation of app to e.g. 32 bit os, by using dlopen instead of linking libraries at compile time
+fontique = { version = "0.7.0", default-features = false, features = ["fontconfig-dlopen"] }
+
 # Try to use only needed features from https://github.com/slint-ui/slint/blob/master/api/rs/slint/Cargo.toml#L23-L31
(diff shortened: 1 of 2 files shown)


krokiet/Cargo.toml
```diff
@@ -35,7 +35,7 @@ rust-embed = { version = "8.5", features = ["debug-embed"] }
 # Try to use only needed features from https://github.com/slint-ui/slint/blob/master/api/rs/slint/Cargo.toml#L23-L31
 #slint = { path = "/home/rafal/test/slint/api/rs/slint/", default-features = false, features = ["std",
 #slint = { git = "https://github.com/slint-ui/slint.git", default-features = false, features = [
-slint = { version = "=1.14.1", default-features = false, features = [
+slint = { version = "=1.15.0", default-features = false, features = [
```
qarmin: = from version needs to be removed here, and in slint-build

I played a little with this and at the end I forgot to push latest commit, which changes this lines

krokiet/Cargo.toml
```diff
@@ -35,7 +35,7 @@ rust-embed = { version = "8.5", features = ["debug-embed"] }
 # Try to use only needed features from https://github.com/slint-ui/slint/blob/master/api/rs/slint/Cargo.toml#L23-L31
 #slint = { path = "/home/rafal/test/slint/api/rs/slint/", default-features = false, features = ["std",
```
qarmin, line 41: To fix 32 bit CI, this dependency must be added
```
# Easier cross-compilation of app to e.g. 32 bit os, by using dlopen instead of linking libraries at compile time
fontique = { version = "0.7.0", default-features = false, features = ["fontconfig-dlopen"] }
```

PR #1784, opened 2026-02-09
Update czkawka_core.ftl

jan-vandenberg:
I started using czkawka last week -- in Dutch -- and noticed lots of strings were still in English or inconsistent with capitalization.

So this patch:

- Adds lots of missing strings (based on the english language file)
- Translates them to Dutch in a consistent manner


qarmin:
Translations are managed on the Crowdin page - https://crowdin.com/project/czkawka
 - please submit updated translations there.

https://github.com/qarmin/czkawka/blob/master/instructions/Translations.md

PR #1797, opened 2026-02-16
Rename Krokiet app icon

malfisya:
Krokiet icon listed in `io.github.qarmin.krokiet.desktop` as io.github.qarmin.krokiet. Make the the actual app icon file and the one listed in the desktop file the same.


malfisya:
I noticed that you did [the opposite](https://github.com/qarmin/czkawka/commit/71f4ffe9bb80fbb4ea5422a317d4453e292375c2) of this PR before. Do you prefer to change the [desktop file](https://github.com/qarmin/czkawka/blob/a0173ce08c10365c57f938bd5a6b60aefbcc5509/data/io.github.qarmin.krokiet.desktop#L4) instead?

qarmin:
I tried to make flatpak work, and this commit was invalid, because id should have 4 components - I changed this in 10.0.0_flatpak branch, but not in master branch yet

PR #1798, opened 2026-02-16
krokiet translation: fix German translation for duplicate files tool

PirMei:
... and revert it to the same as it was in in czkawka.

Background:  
"Dateien duplizieren" in German means "To duplicate/duplicating files" and does not mean "Files, that are duplicates" like the tools intention.  
The name "Gleiche Dateien" as used in czkawka is pretty good for describing the tool's intention, so I used that one.  
A more literal translation would be "Duplizierte Dateien", but "Gleiche Dateien" is probably better for a more diverse target audience.

If there are better ways to help with translation, I was unable to find those :/


qarmin:
Please change this on crowdin page, which is used to manage translations - https://crowdin.com/project/czkawka/de

PirMei:
Sorry, I didn't see this in the instructions folder.  
Might be worthwhile to advertise this in a CONTRIBUTING.md in the project's root.

PR #1809, opened 2026-02-22
Add custom select popup, respect generating thumbnails in similar videos, show file dialogs in separate thread

### LINKED ISSUE #1761: krokiet lacks the select custom option of czkawka_gui_gtk. Please add it.
krokiet lacks the select custom option of czkawka_gui_gtk. Please add it.

Select custom is extremely valueable. Let's say I am comparing two harddrives and find 10,000 duplicates, In the old czkawka gui gtk I can click select custom and select by the path of the drive I want everything deleted from. And better yet czkawka gui gtk has the default option to never select all the files in a group (just in case there are also duplicates I am unaware of within the same harddrive for example.)

Buuuuuut I end up using krokiet gui more just because krokiet duplicate files GUI has a column for file paths. czkawka duplicates files really ONLY works via select custom and setting up a custom path filter. Because I literally cannot see which path which file is from, they appear as identical duplicates and I have no means to choose which to select in this case.

Below first up is czkawka gui gtk's failing:

Image

no column for the path.

Below is czkawka gui gtk's best feature IMO:

Image

And below krokiet lacks any sort of perform filter based selection option from what I can see here:

Image

### LINKED ISSUE #1789: [Crokiet v11.0.0] Clicking Delete button while file dialog is open freezes
**Bug Description**

**Steps to reproduce:**

1. Scan first
2. Check some files and Move
3. The dialog opened for seleting target directory
4. But Crokiet is not hanged and still accessible
5. Click Delete
6. Freezes

```
<!--
Add terminal output only if needed - if there are some errors or warnings or you have performance/freeze issues.  
Very helpful in this situation will be logs from czkawka run with RUST_LOG environment variable set e.g. 
`RUST_LOG=debug ./czkawka` or `flatpak run --env=RUST_LOG=debug com.github.qarmin.czkawka` if you use flatpak, which will print more detailed info about executed function.
-->

<details>
<summary>Debug log</summary>

# UNCOMMENT DETAILS AND PUT LOGS HERE

</details>
```

**System**

- Czkawka/Krokiet version: 11.0.0 gui
- OS version: Windows vNext
- Installation method: github binaries

[krokiet (2).log](https://github.com/user-attachments/files/25314843/krokiet.2.log)

qarmin:
Fixes #1789
Fixes #1761

Screenshot From 2026-02-22 22-13-27

logo

logo_text


eddified:
Awesome! Can't wait to use this. I will check out the nightly build...

PR #1811, opened 2026-02-23
Fix misleading translation for 'tool_duplicate_files' in German

fttr:
Let's use the label from the old UI which is not misleading.


qarmin:
Please change this on crowdin page, which is used to manage translations - https://crowdin.com/project/czkawka/de

Someone changed German translation, so it could be already changed there

PR #1852, opened 2026-03-24
Optimize CLI: enable LTO, use HashMap for cache, reduce allocations

aecs4u:
## Summary
- Enable `lto = "fat"` and `codegen-units = 1` in `[profile.release]` for 5-15% runtime improvement on CPU-bound operations
- Replace `BTreeMap<String, T>` with `HashMap<String, T>` across cache internals and all tool modules — these maps are only used for lookups (O(1) vs O(log n)), ordering is never needed
- Remove unnecessary `String`-allocating sort in `dir_traversal` — sort by `PathBuf` directly instead of `to_string_lossy().to_string()`
- Optimize `diff_loaded_and_prechecked_files`: linear scan for small groups (<4), `HashMap` for larger ones (resolves TODO)
- Move sequential filter into parallel iterator in `similar_images` (resolves TODO at core.rs:301)
- Switch CLI progress channel from `unbounded()` to `bounded(256)`

## Test plan
- [x] `cargo check` passes for all crates
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] All 303 tests pass (0 failures)
- [ ] Benchmark on large directory (10k+ files) to measure improvement
- [ ] Verify cache round-trip works correctly (run twice, second uses cache)


aecs4u:
Please review and accept if applicable

qarmin:
BTreeMap ensures stable results, as can be seen in the tests that otherwise fail.

HashMap was used in the past, but it was replaced with BTreeMap to avoid having to sort results later, so the GUI receives them already sorted (at least to some extent).

As for LTO, this is also intentional - enabling it by default would significantly increase compilation time. Instead, the relevant lines are only enabled in CI, so during development I don’t have to deal with longer build times.

aecs4u:
Thank you.
I believe the migration to HashMap was performed only for structures that do NOT require ordeing.
What are the tests you are refering to?

PR #1888, opened 2026-03-27
Image compare in Krokiet, RMB context menu, changed AV1 codec

### LINKED ISSUE #92: Czkawka GUI - Add possibility to open folder in file browser
When going through a file list it would be nice to be able to right click and then "Open in File Browser" to examine a folder during file clean-up.

### LINKED ISSUE #684: Open directory with Czkawka
I use Czkawka 4.0.0 in Kubuntu Linux.

I would like to open a directory with Czkawka directly from the file manager (Dolphin in my case) using the context menu. By doing so, it seems the program tries to launch but it does nothing.

Thank you.

### LINKED ISSUE #1114: Context menu with open file/open location (especially for defect files)
**Feature Description**
I would like to right-click on a file and open it with my default e.g. image viewer.
Also, maybe a "Open file location" to open the directory (in the ideal case with the file being selected by default) would be helpful.

**Background:**
Especially for the defect files search, I would like to double-check whether the file is _really_ defect or some app can still open it etc. For others, I might want to cross-check the directory context etc.

Generally though, you could make more use of the context menu there.

### LINKED ISSUE #1805: Video Optimizer AV1 single threaded?, and it hangs on encode.
**Bug Description**
Encoding video into AV1 is very slow, the progress bar doesn't move, "Optimized 0/1" no progress on current or all stages, ffmpeg is seen doing maybe 20% cpu in task manager.
**Steps to reproduce:**

Tested using very short twitter videos, 5-20MB which should encode in handbrake AV1 in less than a minute.
Image

FFmpeg Full 8.0.1 from chocolatey

```
<!--
2026-02-18 17:29:40.455 [WARN] [czkawka_core::common::process_utils] (148113) process_utils.rs:96 - Command is still running after 50 seconds, for command: "ffmpeg" "-i" "L:\\ORICO-Y20-2TB\\2025-07-29_stuff_1950326558515687470_JOgi6-JX4MPBg7gC.mp4" "-nostdin" "-c:v" "libaom-av1" "-crf" "23" "-c:a" "copy" "-y" "L:\\ORICO-Y20-2TB\\2025-07-29_stuff_stuff_JOgi6-JX4MPBg7gC.czkawka_optimized.mp4"
2026-02-18 17:33:00.514 [WARN] [czkawka_core::common::process_utils] (148113) process_utils.rs:96 - Command is still running after 250 seconds, for command: "ffmpeg" "-i" "L:\\ORICO-Y20-2TB\\2025-07-29_stuff_JX4MPBg7gC.mp4" "-nostdin" "-c:v" "libaom-av1" "-crf" "23" "-c:a" "copy" "-y" "L:\\ORICO-Y20-2TB\\2025-07-29_stuff_JX4MPBg7gC.czkawka_optimized.mp4"
-->


```

**System**

2026-02-17 09:07:50.514 [INFO] [czkawka_core::common::logger] (1) logger.rs:176 - Krokiet version: 11.0.0(8bd6fb8470 O 2026-02-14), release mode, rust 1.92.0 (2025-12-08), os Windows 10.0.26200 (x86_64 64-bit), 20 cpu/threads, features(0): [], app cpu version: x86-64-v1 (SSE2), os cpu version: x86-64-v3 (AVX2)

- Czkawka/Krokiet version: 
- OS version: 
- Installation method:

qarmin:
Fixes #1805
Fixes #1114
Fixes #684
Fixes #92

- Using for now unstable slint 1.16, which fixes some issues with very slow scrolling in Cedinia(still is slow, but now it)
- Added image comparator for Krokiet
- Added context menu at right click
- Changed Av1 codec to libsvtav1
- Updated translations and added Hindi, Indonesian and Vietnamese languages
- FFmpeg custom command  + noise reduction
- Adding Krokiet.exe into windows zip folder

https://github.com/user-attachments/assets/5243c36b-9e0f-41cf-ab88-73d41877bcc8


skaffi:
That's really great! While the thumbnails have been incredibly useful (including the more recent video thumbnails - thank you for adding that, too), any time there's been any doubt about a pair or a group of similar pictures or images, I almost always need to put down Krokiet, and rely on other tools to make a comparison instead. I can easily imagine that this new feature will allow me to stay within Krokiet a lot more often, and so keep a better flow.

However, I'd like to make one additional feature request, which builds on top of what you've added here. The "side-by-side" view is definitely nice to determine immediate differences at a glance, and the "swap" view will be useful for making a more in-depth comparison. But, I'd like to suggest a third view, in addition - I'll call it a "sliding" view, for lack of a better term. It's the type of image comparison method that's probably the most commonly used on the web in posts and articles of all kinds, be they online news, enthusiast magazines, blogs of all sorts, and so on.

Basically, the two pictures are on top of each other, and there's a vertical line going down the middle, with each picture always only being visible on their own, separate side of the vertical split. You're free to drag the vertical line either left or right, at your own pace, and to change directions as you like, obscuring more of one picture while revealing more of the other.

This method can be really useful, if you need to make a detailed comparison between two pictures. So can the "swap" view, but while there are some types of differences that will immediately stand out when swapping back and forth, there are others that require you to more slowly and methodically visually scan for changes. A "slide" view will be really useful here - not least because repeatedly flashing two different pictures before your eyes can quickly become quite a strain, whereas sliding to reveal and conceal each picture at your own pace should be easy on the eyes, even if you go at it for hours at a time!

Sorry for being a bit verbose. Oh, do I need to raise this feature request as an issue, too, or will this suffice?

qarmin:
Not sure if I understand the feature correctly or if you used the latest version, but I think this was already implemented in one of the next PRs


https://github.com/user-attachments/assets/f5f37c78-1aa6-4fa0-a5ae-5c0c259babcb

PR #1915, opened 2026-04-20
Added `invert selection in group` feature to `krokiet` UI.

kater162:
This change adds a new option to the `krokiet` UI that allows the user to invert the selections of items within a group of duplicate files only if one or more items within the group are selected.

Disclaimer: most of the changes were created using AI coding agents. I have personally built and tested the changed in my Macbook as I needed this feature to sort over 100k duplicate files.

### DIFF
--- a/krokiet/src/connect_select/mod.rs
+++ b/krokiet/src/connect_select/mod.rs
@@ -404,22 +404,24 @@ mod tests {
     fn invert_selection_in_group_only_changes_groups_with_selection() {
         let mut model = get_model_vec(7);
         model[1].header_row = true; // Group 1 header
-        // Group 1 (indices 0): items 0 - no selection initially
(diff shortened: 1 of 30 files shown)


krokiet/src/connect_select/mod.rs
```diff
@@ -110,6 +112,7 @@ pub(crate) fn set_select_buttons(app: &MainWindow) {
             SelectMode::SelectLongestPath,
         ],
         ActiveTab::SimilarImages => vec![
+            SelectMode::InvertSelectionInGroup,
```
qarmin, line 115: This name not really correspond what it does, this is really "InvertSelectionInGroupsThatHaveAtLeastOneItemSelected"(and original Invert Selection also sometimes inverts selection in groups) - but of course it is too long for text in popup, so well, it is possible that I will change this in future, but for now it can stay

Issue #1321, opened 2024-08-08
File Moving Operation Alters Original Mtime


Neher-Veem:
### **Bug Description**
I might be a bit cautious, but every time I search for duplicate files, I follow a process: I first obtain the list of candidates, select the ones I want to delete, and then move them to another directory within the same filesystem. Afterward, I manually review them and perform another search using the original directory as the "reference" and the new one as the "normal."
**I was surprised to find that Czkawka appears to copy, save, or otherwise touch the original files, rather than simply moving them. Moving a file within the same filesystem is an atomic operation that should not alter the mtime.**

### **Steps to reproduce:**

1. Create a copy of any file using the following command:
`rsync -avhtU /my_dir/my_file /my_dir/copy_of_my_file`
This preserves the timestamps as much as rsync allows in Linux.
2. Search for duplicates with Czkawka. When you get the results, select copy_of_my_file, check it, and then move it to a different location using the Move button.
3. Open a terminal and run:
`stat /my_dir/my_file /some_other_dir/copy_of_my_file` 
Review the output, paying special attention to lines starting with "Modify:" and also note that the Inode number has changed.

### ** Expected behavior**
You can either use an atomic move operation of the filesystem to avoid altering the mtime, or use rsync to preserve the mtime and other attributes. Alternatively, in the worst case, please make it clear that moving will involve copying and then deleting the original file, which will result in the loss of the mtime and potentially other attributes.

### **System**
- Czkawka Version 7.0.0 - 19.02.2024r 
- OS version: Linux Mint 22
- Installation method: Binary, AppImage and Flatpak. All of them showed the same behavior

wirwolf:
Have the same problem(((

PR #1932, opened 2026-05-01
Preserve file attributes during move/copy operations

wirwolf:
## Problem

When files are moved from Czkawka/Krokiet using the move action, moved files receive the current modification time instead of keeping their original metadata. This is especially problematic for photo and video collections, where `mtime` is commonly used for sorting and organization.

Reported in: #1321

## What was changed

The original approach of setting `copy_attributes = true` on `fs_extra` options was not possible — `fs_extra` 1.3.0 does not have this field. `fs_extra` has been removed from both crates and replaced with a stdlib-based implementation that correctly preserves attributes.

### `czkawka_gui/src/connect_things/connect_button_move.rs`

- Replaced `fs_extra::dir::move_dir` and `fs_extra::file::move_file` with a custom `move_item_preserving_attrs` function.
- On the same filesystem, uses `std::fs::rename` — atomic, preserves all attributes with zero overhead.
- On cross-device moves, falls back to `copy_item_preserving_attrs` + delete, which restores permissions and timestamps (`mtime`, `atime`) after copying.

### `krokiet/src/file_actions/connect_move.rs`

- Replaced `fs_extra::dir::copy` and `fs_extra::file::copy` with a custom `copy_with_attrs` function.
- For files: uses `std::fs::copy` (preserves permissions on Unix) followed by `std::fs::FileTimes` to restore `mtime` and `atime`.
- For directories: recursively applies the same logic to every entry.

### `czkawka_gui/Cargo.toml` / `krokiet/Cargo.toml`

- Removed `fs_extra` dependency from both crates.

## Expected behavior

When selected duplicate files or folders are moved or copied, their file attributes are preserved:

- modification time (`mtime`) stays unchanged;
- access time (`atime`) stays unchanged;
- file permissions are preserved.

## Why this matters

Changing modification time during move operations breaks workflows where users rely on filesystem timestamps, for example:

- organizing photos and videos by date;
- sorting media collections;
- preserving imported file metadata;
- moving duplicates to a quarantine/review directory before deletion.

### DIFF
--- a/czkawka_gui/src/connect_things/connect_button_move.rs
+++ b/czkawka_gui/src/connect_things/connect_button_move.rs
@@ -166,11 +166,7 @@ fn move_item_preserving_attrs(src: &Path, dst: &Path) -> std::io::Result<()> {
         return Ok(());
     }
     copy_item_preserving_attrs(src, dst)?;
-    if src.is_dir() {
-        std::fs::remove_dir_all(src)
-    } else {
-        std::fs::remove_file(src)
-    }
+    if src.is_dir() { std::fs::remove_dir_all(src) } else { std::fs::remove_file(src) }
 }
 
 fn copy_item_preserving_attrs(src: &Path, dst: &Path) -> std::io::Result<()> {
@@ -183,9 +179,7 @@ fn copy_item_preserving_attrs(src: &Path, dst: &Path) -> std::io::Result<()> {
         }
     } else {
         std::fs::copy(src, dst)?;
-        let times = std::fs::FileTimes::new()
-            .set_accessed(metadata.accessed()?)
-            .set_modified(metadata.modified()?);
+        let times = std::fs::FileTimes::new().set_accessed(metadata.accessed()?).set_modified(metadata.modified()?);
         std::fs::OpenOptions::new().write(true).open(dst)?.set_times(times)?;
     }
     std::fs::set_permissions(dst, metadata.permissions())?;
--- a/czkawka_gui/src/connect_things/connect_button_move.rs
+++ b/czkawka_gui/src/connect_things/connect_button_move.rs
@@ -1,6 +1,5 @@
 use std::path::Path;
 
-use fs_extra::dir::CopyOptions;
 use gtk4::prelude::*;
 use gtk4::{ResponseType, TreePath};
 use log::debug;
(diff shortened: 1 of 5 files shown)


KunalScriptz:
Great work @wirwolf

czkawka_gui/src/connect_things/connect_button_move.rs
```diff
@@ -162,3 +160,28 @@ fn move_files_common(
...
+
+fn move_item_preserving_attrs(src: &Path, dst: &Path) -> std::io::Result<()> {
+    if std::fs::rename(src, dst).is_ok() {
+        return Ok(());
+    }
+    copy_item_preserving_attrs(src, dst)?;
+    if src.is_dir() { std::fs::remove_dir_all(src) } else { std::fs::remove_file(src) }
+}
+
+fn copy_item_preserving_attrs(src: &Path, dst: &Path) -> std::io::Result<()> {
```
qarmin, line 172: copy/move functions should be moved to czkawka_core, because this is thing duplicated between all gui backends(and in future maybe even CLI will add this)

czkawka_gui/src/connect_things/connect_button_move.rs
```diff
@@ -162,3 +160,28 @@ fn move_files_common(
...
+    if metadata.is_dir() {
+        std::fs::create_dir_all(dst)?;
+        for entry in std::fs::read_dir(src)? {
+            let entry = entry?;
+            copy_item_preserving_attrs(&entry.path(), &dst.join(entry.file_name()))?;
+        }
+    } else {
+        std::fs::copy(src, dst)?;
+        let times = std::fs::FileTimes::new().set_accessed(metadata.accessed()?).set_modified(metadata.modified()?);
+        std::fs::OpenOptions::new().write(true).open(dst)?.set_times(times)?;
```
qarmin, line 183: This probably for readonly files will throw an error - don't know this for sure, so test, testing this potential problem should be added to PR

czkawka_gui/src/connect_things/connect_button_move.rs
```diff
@@ -162,3 +160,28 @@ fn move_files_common(
 
     text_view_errors.buffer().set_text(messages.as_str());
 }
+
+fn move_item_preserving_attrs(src: &Path, dst: &Path) -> std::io::Result<()> {
+    if std::fs::rename(src, dst).is_ok() {
```
qarmin, line 165: Not for all rename errors, we should use copy fallback.

I think, that only in case of ErrorKind::CrossesDevices, this should try to move file/folder with alternative method

qarmin (changes requested):

luzpaz:
Hi @wirwolf, are you still planning to work on this ?

wirwolf:
> Hi @wirwolf, are you still planning to work on this ?

Hey. Maybe a little bit later. I build your project locally, run it on my NAS and find all issues in my data. If you ask about comments for code formatting - i add finishing this MR to my to-do list

PS: Also, if it's interesting, I run this project in a Docker container to run on a remote server and control the application on a browser. Its very comfortable

PR #1933, opened 2026-05-01
Add 'heif' feature to default dependencies

wirwolf:
For recognizing photos from iphone


qarmin:
This is a very bad idea. The main reason this feature is not enabled by default is to keep the build process simple. Adding it introduces an extra dependency and breaks out-of-the-box support on many platforms.

The second issue is that this feature can already be enabled if needed, and I also provide prebuilt binaries with HEIF support.

PR #1938, opened 2026-05-11
GTK GUI: Replace select popover with a structured Select Dialog

wirwolf:
The "Select" popover in the GTK GUI was growing into an unmanageable flat list of buttons — each new combination of criterion (date / path / size) and condition (same size / same path / longest path…) required its own dedicated button, visibility logic, and i18n string.

This PR replaces that approach with a compact **Select Dialog** window that exposes the same functionality through composable dropdowns and radio groups, and adds several new selection modes that would have been impractical to fit in the old popover.

### New: Select Dialog (`gui_select_dialog.rs`, `connect_select_dialog.rs`)

A modal dialog (`gtk4::Window`) opened by the existing "Select" button. It is structured into frames:

| Frame | Options |
|---|---|
| **Action** | Select all except… / Select one… / Select All / Unselect All / Reverse / Mark groups with same size / Custom select / Custom unselect |
| **Criterion** | Date / Path length / Size |
| **Direction** | Oldest–Newest / Shortest–Longest / Smallest–Biggest (shown depending on Criterion) |
| **Path filter** | Any / Longest only / Shortest only (shown for Action=Select one + Criterion=Date) |
| **Conditions** | None / Same size only / Same path only |
| **Custom filter** | Filter by name / path / regex with case-sensitivity and "all in group" options |

Frames are shown or hidden dynamically based on the selected Action and Criterion, so only relevant options are presented at any time.

### New selection modes added

- **Select one oldest/newest (same size)** — selects one file per group by date, but only applies to groups where all duplicates share the same file size.
- **Select one oldest/newest (same path)** — same, but constrained to groups where all duplicates reside in the same directory.
- **Select one oldest/newest (longest/shortest path)** — among duplicates with the longest (or shortest) path in the group, picks the one with the extreme modification date.
- **Select one longest/shortest path (same size)** — picks the file with the extreme path length, but only in groups where all files are the same size.
- **Mark groups with same size** — highlights (marks) duplicate groups where every file in the group has the same size.

### Refactoring: `connect_popovers_select.rs`

All selection functions were converted from private `popover_*()` (accepted `&gtk4::Popover`, called `popdown()` at the end) to public `exec_*()` (accept only `&SubView`). This separates business logic from UI teardown, making the functions reusable from both the old popover buttons and the new dialog.

### Refactoring: `connect_button_select.rs`

The `show_required_popovers()` function (~100 lines of manual `set_visible()` calls for every individual button) is removed. The "Select" button now calls `select_dialog.set_available_modes(sv.nb_object.available_modes)` and shows the dialog. Mode availability (which frames are enabled) is handled centrally inside `GuiSelectDialog`.

### i18n

All new strings are provided in **English** (`en/czkawka_gui.ftl`) and **Russian** (`ru/czkawka_gui.ftl`).

## Motivation

Before this change, adding a single new selection criterion required:
1. A new button widget in `popover_select.ui`
2. A new field in `GuiSelectPopovers`
3. A new `set_visible()` branch in `show_required_popovers()`
4. A new `connect_clicked` handler
5. Two new i18n strings

With the Select Dialog, a new criterion is a new radio button in the dialog and a branch in `dispatch_apply()`. The popover itself becomes much simpler and the new dialog scales cleanly as functionality grows.


qarmin:
12.0 is the latest version of Czkawka GUI. I already mentioned that I only accept bugfixes for it, because maintaining it, testing changes, and keeping everything stable requires too much effort.  Because of that, I decided not to merge these kinds of changes anymore. 
 
Feel free to create and distribute a fork with them included.

Issue #1225, opened 2024-02-24
"Presets" does not work correctly


flansuse:
**Summary**
"Added directories" are not saved when switching between presets

**STEPS TO REPRODUCE**
1. Select *Preset 1*
2. Add a few directories under "Included Directories"
3. Click "Save"  (it will flash the message "Saved preset 1")
4. Choose *another* preset
5. Now select *Preset 1* again
6. **The added directories are gone!**

**System**
- Krokiet version: 7.0.0
- Arch Linux
- Installation method:  Launching precompiled binary `linux_krokiet_libraw_heif`

qarmin:
Eh...
I fixed it multiple times but looks that still I messed with some properties,

To fix this, just click "Load" an this will load proper preset.
Changing presets looks that is off by one - when you save preset 2, you can see that changes are visible on preset 1 after moving, but clicking load will fix them

PR #1940, opened 2026-05-12
Persist preset names in Krokiet on Save Current Preset click

johntrandall:
## Problem

Krokiet's **Save Current Preset** button writes the per-preset config file (`config_preset_N.json`) but does not write the base settings file (`config_general.json`). Preset names live in `BasicSettings` and are only flushed on clean app exit via `save_all_settings_to_file()`. On a non-clean exit (window close that returns non-`Ok`, logger that did not flush, force quit), preset renames silently vanish on next launch.

## Repro (Krokiet 11.0.1, macOS 15.7)

1. Open Krokiet
2. Settings tab → click **Edit name** next to a preset, type a new name, click **Save**
3. Click **Save Current Preset**
4. Quit the app
5. Relaunch
6. Preset name has reverted to the default ("Preset 1", etc.)

## Fix

`on_save_current_preset` now calls `save_base_settings_to_file` in addition to the existing per-preset write, so the Save button matches user expectations — Save means save everything visible, including any preset rename just made through the Edit name dialog.

Adds a round-trip unit test for `BasicSettings` in `krokiet/src/settings/mod.rs` to lock in the data path the new call exercises.

## Verification

- `cargo test -p krokiet` — passes (new test `basic_settings_round_trips_renamed_preset_names`)
- `cargo +nightly fmt` + `cargo clippy --all-targets -- -D warnings` — clean
- Manual end-to-end on macOS 15.7 in a clean Tart VM:
  - Rename `Preset 1` → `TEST-FIX-VERIFY 1` in the Edit name dialog
  - Click Save Current Preset → `config_general.json` mtime updates, `preset_names[0] = "TEST-FIX-VERIFY 1"`
  - `pkill -9` (worst-case force-quit) → file content survives
  - Relaunch → Settings combo box shows `TEST-FIX-VERIFY 1`
  - Without the fix, the same sequence leaves `config_general.json` unchanged (still defaults).

## Related

- #1225 (closed 2024-02-24) — "'Presets' does not work correctly" was a previous preset-persistence bug, fixed for the per-preset config path. The base settings file (where preset names live) was not covered by that fix.

## Scope

Intentionally minimal: one new call inside the existing callback, plus one unit test. The narrower follow-up — saving base settings directly from the Edit name dialog's own Save button, so rename-then-force-quit-without-clicking-Save-Current-Preset doesn't lose the rename — is out of scope here. That would require a new `Callabler` callback and wiring it from `krokiet/ui/settings_list.slint`.

### DIFF
--- a/krokiet/src/settings/mod.rs
+++ b/krokiet/src/settings/mod.rs
@@ -58,6 +58,10 @@ pub(crate) fn connect_changing_settings_preset(app: &MainWindow) {
                 error!("Failed to save preset - {e}");
             }
         }
+        // Also persist base settings (preset names, default preset, window prefs).
+        // Without this, renaming a preset only takes effect after a clean app exit,
+        // and is lost if the on-exit save path doesn't run.
+        save_base_settings_to_file(&app, current_item);
     });
     let a = app.as_weak();
     app.global::<Callabler>().on_reset_current_preset(move || {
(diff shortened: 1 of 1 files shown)


krokiet/src/settings/mod.rs
```diff
@@ -975,3 +979,33 @@ pub(crate) fn collect_base_settings(app: &MainWindow) -> BasicSettings {
...
+    use std::path::PathBuf;
+
+    use super::{load_data_from_file, save_data_to_file};
+    use crate::settings::model::BasicSettings;
+
+    // The bug fixed in `on_save_current_preset` (writing base settings on Save click) relies
+    // on this round-trip: edit preset_names in memory → save_data_to_file → load_data_from_file
+    // must return the edited names. Without it, renaming a preset cannot survive a restart.
+    #[test]
+    fn basic_settings_round_trips_renamed_preset_names() {
```
qarmin, line 994: This test don't do anything except saving data to file and reading it via fs and serde options, so it doesn't check almost any logic that could regress in future(this items are quite hide to broke) and this test isn't related in any way with this PR - please remove it.

krokiet/src/settings/mod.rs
```diff
@@ -58,6 +58,10 @@ pub(crate) fn connect_changing_settings_preset(app: &MainWindow) {
                 error!("Failed to save preset - {e}");
             }
         }
+        // Also persist base settings (preset names, default preset, window prefs).
+        // Without this, renaming a preset only takes effect after a clean app exit,
```
qarmin, line 62: This code doesn't need a comment(it is easy to deduce, what this do) so comment needs to be removed

PR #1941, opened 2026-05-12
krokiet: draggable splitter for Included/Excluded Paths panel

### LINKED ISSUE #1792: Krokiet - Resizable splitter for Included/Excluded Paths panel
**Feature Description**

It would be great to have a draggable splitter/divider handle on the horizontal bar that separates the main results area from the "Included Paths" and "Excluded Paths" panel at the bottom of the window.

**Current behavior**

The bottom panel has a fixed height. There is no way to resize it — you cannot see more paths if the list grows long.

**Desired behavior**

Clicking and dragging the divider bar upward (or downward) should resize the panel, giving more or less vertical space to the paths section. This is similar to split-pane behavior common in file managers and IDEs.

**Why it's useful**

When many included/excluded paths are configured, the current fixed-height panel truncates the list and requires scrolling. Being able to drag the panel taller would make it much easier to manage paths.

**Environment** 
  - Krokiet 11.0.0
  - Arch / Gnome / Wayland

Image

johntrandall:
## Summary

Adds a draggable splitter handle above the Included/Excluded Paths panel in krokiet, so the user can resize the bottom panel by dragging. Height persists across launches.

Closes #1792.

## Changes

- **`krokiet/src/settings/model.rs`** — new `bottom_panel_height: u32` field on `BasicSettings` (`#[serde(default)]`, default 150, clamp 60..1080*4). Round-trip + missing-field tests added. Older config files (no `bottom_panel_height` key) deserialize cleanly via serde default.
- **`krokiet/src/settings/mod.rs`** — load/clamp in `set_base_settings_to_gui`, save in `collect_base_settings`. Mirrors the existing `window_height` plumbing.
- **`krokiet/ui/settings.slint`** — new `Settings.bottom_panel_height: length` global (default 150px).
- **`krokiet/ui/bottom_panel.slint`** — `BottomPanel` `min-height` floor relaxed from 150px to 60px so the user can drag small.
- **`krokiet/ui/main_window.slint`** — 6px draggable `Rectangle` splitter handle with a `TouchArea` (mouse-cursor: row-resize) above the bottom panel; `BottomPanel.height` bound to `Settings.bottom_panel_height` with clamp `60px..root.height - 200px`.

## Drag math note

Each `moved` event applies `(mouse-y - pressed-y)` to the **live** `Settings.bottom_panel_height`, not to a press-time snapshot. The snapshot pattern is unstable here because the `TouchArea` itself translates with the splitter as the panel grows; `mouse-y` in the new TouchArea-local frame returns to `pressed-y` each event, zeroing the delta. The "increment from current" idiom is stable because `Δmouse-y` matches the layout-induced `Δ-TA-origin` one-to-one. This matches the canonical splitter pattern from <https://docs.slint.dev/latest/docs/slint/guide/development/custom-controls>.

## Layout note

`BottomPanel` size is controlled via an explicit `height:` binding (per Slint positioning-and-layouts: elements with a fixed width/height are not resized by the layout). Pinning via `min-height` + `max-height` was tried but the constraint propagates to the parent `VerticalBox`'s min-height sum and caused overflow that broke hit-testing on the sibling `LeftSidePanel` tabs.

## Test plan

- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy -p krokiet --all-targets -- -D warnings` — clean
- [x] `bash misc/run_checks.sh` — clean (no unused slint imports / translations / callbacks / settings properties; `bottom_panel_height` is referenced in main_window.slint)
- [x] `cargo test -p krokiet --bin krokiet settings::model` — 4/4 pass (defaults, missing-field deserialization, round-trip)
- [x] Manual verification in a macOS Tart VM via Screen Sharing: hover shows row-resize cursor; drag visibly resizes the panel; clamps at floor and ceiling; sidebar tabs remain clickable; height persists across launches.

### DIFF
--- a/krokiet/src/settings/model.rs
+++ b/krokiet/src/settings/model.rs
@@ -32,6 +32,7 @@ pub const DEFAULT_MINIMAL_FRAGMENT_DURATION_VALUE: f32 = 5.0;
 pub const MAX_HASH_SIZE: f32 = 40.0;
 pub const DEFAULT_WINDOW_WIDTH: u32 = 800;
 pub const DEFAULT_WINDOW_HEIGHT: u32 = 600;
+pub const DEFAULT_BOTTOM_PANEL_HEIGHT: u32 = 150;
 pub const DEFAULT_MIN_VIDEO_THUMBNAIL_POSITION_PERCENT: u8 = 1;
 pub const DEFAULT_MAX_VIDEO_THUMBNAIL_POSITION_PERCENT: u8 = 99;
 
@@ -303,6 +304,8 @@ pub struct BasicSettings {
     pub window_width: u32,
     #[serde(default = "default_window_height")]
     pub window_height: u32,
+    #[serde(default = "default_bottom_panel_height")]
+    pub bottom_panel_height: u32,
     #[serde(default = "detect_language")]
     pub language: String,
     #[serde(default = "ttrue")]
@@ -492,6 +495,9 @@ pub(crate) fn default_window_width() -> u32 {
 pub(crate) fn default_window_height() -> u32 {
     DEFAULT_WINDOW_HEIGHT
 }
+pub(crate) fn default_bottom_panel_height() -> u32 {
+    DEFAULT_BOTTOM_PANEL_HEIGHT
+}
 pub(crate) fn default_video_optimizer_mode() -> String {
     "transcode".to_string()
 }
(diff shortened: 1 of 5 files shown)


krokiet/ui/main_window.slint
```diff
@@ -209,9 +209,46 @@ export component MainWindow inherits Window {
...
+        // returns to pressed-y each frame, zeroing the delta and reverting
+        // the change. The "increment from current value" idiom is stable
+        // through that feedback because Δmouse-y equals the layout-induced
+        // Δ-TA-origin, so the same delta keeps applying each event.
+        Rectangle {
+            vertical-stretch: 0.0;
+            height: 6px;
+            background: splitter_ta.has-hover || splitter_ta.pressed
+                ? Palette.accent-background
+                : Palette.alternate-background;
```
qarmin, line 228: At least for now, this should not change color on hover, because that is inconsistent with the other splitters in the app

krokiet/ui/main_window.slint
```diff
@@ -209,9 +209,46 @@ export component MainWindow inherits Window {
...
+        // panel height as the base. Capturing a press-time snapshot does NOT
+        // work here — the TouchArea itself moves vertically with the splitter
+        // as the panel grows, so mouse-y in the new TouchArea-local frame
+        // returns to pressed-y each frame, zeroing the delta and reverting
+        // the change. The "increment from current value" idiom is stable
+        // through that feedback because Δmouse-y equals the layout-induced
+        // Δ-TA-origin, so the same delta keeps applying each event.
+        Rectangle {
+            vertical-stretch: 0.0;
+            height: 6px;
```
qarmin, line 225: This looks a bit too big to me. Maybe 3-4px would be enough?

krokiet/src/settings/model.rs
```diff
@@ -577,3 +583,40 @@ impl Default for SavedCustomSelectTabState {
...
+    fn test_bottom_panel_height_round_trip() {
+        let settings = BasicSettings {
+            bottom_panel_height: 275,
+            ..Default::default()
+        };
+        let json = serde_json::to_string(&settings).expect("serialize");
+        let restored: BasicSettings = serde_json::from_str(&json).expect("deserialize");
+        assert_eq!(restored.bottom_panel_height, 275);
+    }
+}
```
qarmin, line 622: All these tests are completely useless - please remove them all

krokiet/ui/main_window.slint
```diff
@@ -209,9 +209,46 @@ export component MainWindow inherits Window {
...
+        // Uses the incremental Slint pattern from docs.slint.dev: each `moved`
+        // event adjusts the target by (mouse-y - pressed-y), using the LIVE
+        // panel height as the base. Capturing a press-time snapshot does NOT
+        // work here — the TouchArea itself moves vertically with the splitter
+        // as the panel grows, so mouse-y in the new TouchArea-local frame
+        // returns to pressed-y each frame, zeroing the delta and reverting
+        // the change. The "increment from current value" idiom is stable
+        // through that feedback because Δmouse-y equals the layout-induced
+        // Δ-TA-origin, so the same delta keeps applying each event.
+        Rectangle {
```
qarmin, line 223: Splitter is still visible and clickable, when lower panel is hidden.
Also moving it when bottom panel is hidden and then showing bottom panel, will move position of splitter to strange location

Issue #943, opened 2023-03-11
opus audio file support


themarcnet:
Hi,  First of all, great app - does what I want, when I want,  with no hassle....apart from one thing....

Would it be possible to add ".opus" as an audio format (extension is also .opus) that you scan the tags for.  Opus files use the same tagging as ogg files, which is no suprise since opus and ogg come from the same developers.    

I have many ogg, m4a, mp3, and opus files, and I would like to get rid of the duplicates (i.e: remove the ogg files that have the same tagging information as the newer opus files)

In fact, I can rename opus files to ogg, and your application can read the tags just fine!  Although it would be really nice for it to just work, without the need for this work-around.

More information about opus here -> https://github.com/xiph/opus

mr-brune:
yes it would be usefull also add fingerprint duplicate finder for opus

AKL55:
Would be nice to add as well.

GitKopfkino:
This feature would be helpful!

DocMAX:
+1

Heavygrass:
+1

Kilohertz77:
Replying to ask the exact same question as the OP did. 

Czkawka can already process Opus files when they have the .ogg extension, so please add .opus to the list of allowed extensions.

I now have a batch file set up that first renames all my .opus files to .ogg, then runs Czkawka, and then renames .ogg back to .opus. With a library of thousands of music files this takes too long and shouldn't be necessary.

csolisr:
> Czkawka can already process Opus files when they have the .ogg extension, so please add .opus to the list of allowed extensions.

From what I see, adding `.opus` to the allowed extensions should be trivial, just a matter of appending ` "opus",` to this line: https://github.com/qarmin/czkawka/blob/f82bd5ddf4e1958651249a2b8e9db38df58eadb7/czkawka_core/src/common/consts.rs#L32

csolisr:
(Also, for documentation purposes, to the following other lines): 

https://github.com/qarmin/czkawka/blob/f82bd5ddf4e1958651249a2b8e9db38df58eadb7/instructions/Instruction.md?plain=1#L538

https://github.com/qarmin/czkawka/blob/f82bd5ddf4e1958651249a2b8e9db38df58eadb7/czkawka_cli/src/commands.rs#L809

Eoin-ONeill-Yokai:
@csolisr Have you tested this and, if so, care to make it a merge request?

[...]

PR #1980, opened 2026-06-27
Allow the identification of duplicate audio files with "opus" extension

Eoin-ONeill-Yokai:
Based on the suggestion by Carlos Solís ( @csolisr ) in a discussion on issue number #943, I've run the code myself and can confirm this is all that was necessary for rudimentary opus file format support.

This should be good enough for a merge as it helped unblock me with my duplicate music library issues.


qarmin:
Opus is not supported in Symphonia, so the constant needs to be split into two separate lists, one for finding music based on tags, and another for finding music based on identical content. I’m planning to release a new version today, so instead of waiting for changes, I will implement this on my own for now.

Eoin-ONeill-Yokai:
Ah ok, that explains some of the inconsistent results of the fingerprint. If I manually generate the fingerprint, I get matches using a python script but cannot replicate that in czkawka. 

Thanks for merging though -- I do have a follow up PR for adding album parsing to tag match scanning so I'll accommodate to the new design before posting.

PR #1982, opened 2026-06-28
Version 12.0.0

qarmin:


skaffi:
This is lovely news! I was just stopping by today by chance, and happened to see it as it was compiling. This looks like it's going to be a really great update - the Nightly build I've been running since mid May was already a really nice upgrade itself.

But, I'm afraid I have something less pleasant to point out. [One of the VirusTotal links](https://www.virustotal.com/gui/file/cee1d000181c6ae8ebb7cb6d955edf4ac10aa0449654af2cfe8ce954df44e539) you posted (or, two out of three links, since you seem to have pasted the same link twice) reports that the analysed package (windows_czkawka_gui_gtk_412.zip) is infected with malware (Win32.Malware.KillAV.P1WUEC@gen aka. killav/p1wuec). This is really concerning. Even if it's only the Windows package that shows signs of malware, I will still caution against downloading any of the other available packages, if a part of your setup has been compromised. Also, a link to a VirusTotal analysis of the packaged Linux binaries were not posted, unlike for the Windows binaries and the Android apk - presumably one of the duplicate (heh) links to the Windows binary analysis was supposed to have been for the Linux binaries instead? But the fact that it's missing, as one of the actually scanned packages gets flagged, is certainly foreboding.

Anyway, hopefully this is just a matter of a false positive. But if any of your systems have been infected, I hope you'll be able to contain it and resolve it swiftly. Please keep the community advised, as best you're able, to prevent the spread of any potential malware, and to hopefully be able to warn anyone who may already have inadvertently downloaded a potentially infected package. Best of luck, and again, hopefully this is just a big nothingburger.

qarmin:
All binaries are built directly by GitHub CI and are validated that they are reproducible. None of them are compiled or modified on my PC.

My release flow is simple
- Create a PR with the new version
- Merge PR into master branch, which automatically builds all binaries and uploads them to the Nightly tag - https://github.com/qarmin/czkawka/releases/tag/Nightly
- Download those artifacts and attach exactly the same files to the final release

So my role is essentially limited to copying the artifacts from one GitHub release to another, without modifying them in any way.

Regarding the VirusTotal links, I added them rather quickly before publishing the release, so some were duplicated and some were missing. I've already fixed that, and all Windows binaries as well as the Cedinia APK now have their own VirusTotal reports. I don't upload VirusTotal reports for Linux or macOS binaries simply because there isn't much demand for them, and adding reports for every build would clutter the release page with 50 of additional links, which would look very bad

VirusTotal indeed reported a few detections
- https://www.virustotal.com/gui/file/fbdc5ced0fefd6f3222cf4920ede2c2c9f6286433a6a96b9c1e8e525c788597e
- https://www.virustotal.com/gui/file/cee1d000181c6ae8ebb7cb6d955edf4ac10aa0449654af2cfe8ce954df44e539
- https://www.virustotal.com/gui/file/6c65904eeb1de5b76a338906bc09bee5a4ddf1c977fff836a6cb3a41023a3dd4

The problem is that these are almost certainly false positives (I would be much more concerned if more than five antivirus engines flagged them), which are unfortunately quite common. Modern antivirus engines rely heavily on heuristic detection, and those heuristics sometimes misclassify legitimate software, so one or two detections look more like noise than a real threat. It also probably doesn't help that the project is written in Rust and that I build release binaries with lto = "fat", which enables very aggressive optimizations and produces binaries that are significantly harder for heuristic engines to analyze accurately.

If you look at older releases, you'll find similar detections there as well

https://www.virustotal.com/gui/file/2f270efff5db9a97aeaab1668b5dfb8c55247ba58711d078976a97a5f25e2103

So for me they are completelly safe, just like any other version of this app

skaffi:
Hi Qarmin,

Thank you for taking the time to make a good, and thorough counterpoint, in such a timely manner.

You know, having seen that you're using GitHub CI, I already knew/assumed that your setup would be more or less as you describe. In addition, your point about modern AVs being incredibly reliant on heuristics - and quite error-prone heuristics, at that - is something I know I was also already aware of. Funny how it is, then, that seeing a little bit of text in red on VirusTotal could seemingly make me forget all that, and instead start shouting "red alert".

What you say is true, and your disposition is well reasoned. You've put my concerns at ease. I'm sorry for blowing things out of proportion - especially if such posts could end up reflecting negatively on the project. I'd be happy to delete my previous comment, or I can let it stand, as you prefer.

But for now, I'm off to take the 12.0.0 incarnation of the great counter-duping warsheep _Krokiet_ out for a spin! Though, with a croissant on the banner, I must confess, as that was the closest thing to a croquette that was available to me in my icon theme of choice. ;)
krokietssant

PR #1995, opened 2026-07-05
Surface rollback rename failure in make_hard_link/make_file_symlink

### LINKED ISSUE #1991: Silent data loss when hardlink/symlink rollback rename fails in make_hard_link/make_file_symlink
## Problem
In `czkawka_core/src/common/fs_ops.rs`, `make_hard_link` and `make_file_symlink` swallow the rollback rename error, so a user's original destination file can be left orphaned at a `.czkawka_tmp` path with no error surfaced.

## Steps
1. Call `make_hard_link(src, dst)` (or `make_file_symlink`) where `src` and `dst` are on different devices/filesystems (e.g. dst on FUSE/rclone mount).
2. `fs::rename(dst, temp)` succeeds (both on dst_dir).
3. `fs::hard_link(src, dst)` fails (CrossesDevices/Unsupported).
4. Recovery `let _ = fs::rename(&temp, dst);` at line 196 (hardlink) / line 247 (symlink) — its error is discarded.

## Expected
If the rollback rename fails, surface it (chain into the returned error) so caller knows dst is now at `<dst>.<rand>.czkawka_tmp`.

## Actual
Caller receives only the original `hard_link`/`symlink` error. Original `dst` file remains at temp path, invisible to user.

## Cite
`czkawka_core/src/common/fs_ops.rs:196` and `:247` — `let _ = fs::rename(&temp, dst);`

## Environment
czkawka_core 12.0.0, rustc 1.94.1, edition 2024 (per `czkawka_core/Cargo.toml`).

Thanks for maintaining qarmin/czkawka!

Nikolai-coder:
## What was broken

`make_hard_link` and `make_file_symlink` in `czkawka_core/src/common/fs_ops.rs` create a link by first renaming the existing destination file aside to a temporary `<dst>.<rand>.czkawka_tmp` path, then creating the link in its place. If link creation fails, they rename the original file back.

The result of that rollback rename was discarded with `let _ = fs::rename(&temp, dst)`. If the rollback rename itself failed (for example the destination directory became read-only, or another process recreated `dst` in the meantime), the caller received only the original link error. It was never told that the user's original file had been left under the temporary `.czkawka_tmp` name instead of at `dst`.

The original file is not deleted, so it stays recoverable at the temporary path, but nothing in the reported error lets the user know where it went.

## Root cause

`fs_ops.rs:196` (hard link) and `fs_ops.rs:247` (symlink): the rollback `fs::rename(&temp, dst)` result was dropped via `let _ = ...`.

## Fix

Extract the rollback into a small helper `recover_from_failed_link(temp, dst, primary_error)`:

- When the rollback rename succeeds, it returns the primary link error unchanged, so the common case behaves exactly as before.
- When the rollback rename fails, it returns an error with the same `ErrorKind` whose message folds in the original failure, the temporary path where the file now lives, and the rollback error, so the user can find and restore the file.

Both call sites now go through it. This mirrors the existing `describe_hardlink_error` helper already in this file.

## Tests

Two deterministic, cross-platform unit tests were added:

- `recover_from_failed_link_returns_primary_error_when_rollback_succeeds` - the rollback restores `dst`, the temp file is gone, and the returned error is the unchanged primary error.
- `recover_from_failed_link_reports_orphaned_file_when_rollback_fails` - with the rollback forced to fail, the returned error keeps the original `ErrorKind` and its message names the primary failure and the orphaned temp path.

## Validation (local)

- `cargo test -p czkawka_core --lib`: 231 passed, 0 failed, 6 ignored.
- `cargo +nightly fmt --check`: the changed file is clean.
- `cargo clippy -p czkawka_core --all-targets`: no errors and no new warnings from this change. (There is one pre-existing `while_let_loop` warning at `fs_ops.rs:55` in `check_if_folder_contains_only_empty_folders`; it is unrelated to this change and left untouched.)

Tested locally on Windows only; not run on Linux or macOS. The code uses only `std::fs`, so the behavior should be the same on those platforms, but I have not verified it there.

## Notes

The combined-failure path (link creation fails and the rollback rename also fails) cannot be reproduced deterministically on a normal filesystem, because the temporary file is in the destination's own directory and the rollback normally always succeeds. For that reason the rollback logic was extracted into a directly testable helper, following the pattern already used for `describe_hardlink_error`.

Closes #1991

### DIFF
--- a/czkawka_core/src/common/fs_ops.rs
+++ b/czkawka_core/src/common/fs_ops.rs
@@ -192,11 +192,26 @@ pub fn make_hard_link<P: AsRef<Path>, Q: AsRef<Path>>(src: P, dst: Q) -> io::Res
             fs::remove_file(&temp)?;
             Ok(())
         }
-        Err(e) => {
-            let _ = fs::rename(&temp, dst);
-            Err(describe_hardlink_error(e, dst))
-        }
+        Err(e) => Err(recover_from_failed_link(&temp, dst, describe_hardlink_error(e, dst))),
     }
+}
+
+// After moving the original `dst` aside to `temp`, the hardlink/symlink creation failed. Move the
+// original back. If that rollback rename also fails, the original file is now orphaned at `temp`;
+// returning only the primary error would hide that (silent data loss, issue #1991), so fold the
+// rollback failure and the temp location into the returned error.
+fn recover_from_failed_link(temp: &Path, dst: &Path, primary_error: io::Error) -> io::Error {
+    let Err(rollback_error) = fs::rename(temp, dst) else {
+        return primary_error;
+    };
+    Error::new(
+        primary_error.kind(),
+        format!(
+            "{primary_error}; original file could not be restored and is now left at \"{}\" (rename back to \"{}\" failed: {rollback_error})",
+            temp.to_string_lossy(),
+            dst.to_string_lossy()
+        ),
+    )
 }
 
 // rclone/network/FUSE mounts and cross-device targets can't hold a hard link; the bare
@@ -243,10 +258,7 @@ pub fn make_file_symlink<P: AsRef<Path>, Q: AsRef<Path>>(src: P, dst: Q) -> io::
             fs::remove_file(&temp)?;
             Ok(())
         }
-        Err(e) => {
-            let _ = fs::rename(&temp, dst);
-            Err(e)
-        }
+        Err(e) => Err(recover_from_failed_link(&temp, dst, e)),
     }
 }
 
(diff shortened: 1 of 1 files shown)


Nikolai-coder:
CI note: the three failing checks are also failing on the exact base commit in `master` (`2f20e2fe5`), as shown in [run 28411936801](https://github.com/qarmin/czkawka/actions/runs/28411936801).

* `linux-all-extra` on x86_64 and ARM64 fail while linking Krokiet because `freetype`/`fontconfig` development libraries are unavailable.
* `linux-tests` fails because `cargo-clippy` is not installed for Rust `1.94.1`.

All checks exercising this change across Windows, macOS and Linux, including the minimum-Rust regression tests and the quality job, passed. No workflow or unrelated code changes are included in this PR.

Issue #1987, opened 2026-07-02
Text corruption when selecting Korean in Cedinia settings


tjdbs020108:
As the title says, when I select Korean as the language in Cedinia, all text becomes garbled.

Is this because the program does not include a Korean font?

In addition, file paths are displayed with the same garbled characters instead of Korean text.

<img width="648" height="1390" alt="Image" src="https://github.com/user-attachments/assets/e5faca22-c494-4dc7-933d-19f7e7316823" />

qarmin:
Which version are you using, 12.0.0 or the 11.0.1 nightly snapshots?

With the latest version, I can't reproduce this problem. Before updating Slint, I had a similar issue in 11.0.1, but it disappeared after upgrading to Slint 1.17.

Fonts are not bundled with the APK because the operating system is expected to provide them.

In https://github.com/qarmin/czkawka/pull/1989 I added an experimental option to bundle some fonts. It doesn't break anything on my device, but it's possible that it also doesn't fix this issue.

Could you check whether the problem is still present with this build?

https://github.com/qarmin/czkawka/actions/runs/28681738448?pr=1989 (At the bottom - this is debug version, a lot of slower and bigger, but should be enough for tests)

https://github.com/qarmin/czkawka/actions/runs/28681738448/artifacts/8074451899 (APK)

tjdbs020108:
If I remember correctly, I installed version 12.0.0 from the latest release.
I uninstalled the app and reinstalled it, but the text is still garbled.
I also installed the debug APK, but the issue persists.
Could this be related to my phone model? I'm using a Samsung Galaxy S23 Ultra.

qarmin:
I tested it on a Samsung A15 alongside my Motorola, and both worked fine.

I'm not sure what could be causing the issue. Samsung devices seem to include CJK fonts, at least on the model I tested, so I assume other models behave similarly. 

I don't think the issue will be fixed unless I can find a device where I can test potential fixes.

<img height="900" alt="Image" src="https://github.com/user-attachments/assets/e5b319d9-6ed4-47d8-b513-a174e17ad591" />

tjdbs020108:
I asked ChatGPT about this issue, so these are only guesses and could be completely wrong. If that's the case, I'm sorry. 🥹

ChatGPT suggested that it might not be an encoding issue, because English and Chinese are displayed correctly, while only Korean characters are rendered as square boxes (□). It also pointed out that even Korean file names are displayed as squares, not just the translated UI text.

[...]

PR #2022, opened 2026-08-01
Fix missing Chinese glyphs in Krokiet

Qianli-Ma:
Krokiet’s Chinese translations were loaded correctly, but rendering depended on a suitable CJK font being installed on the host system. On systems without one, Chinese text appeared as missing glyphs.
This change:
Bundles the official Noto Sans CJK font collection.
Configures it as Krokiet’s default UI font.
Adds the SIL Open Font License and attribution.
Supports both Simplified and Traditional Chinese text, including translated UI strings and Chinese file names.


qarmin:
A 20 MB binary is one of the things I really don't want to have in the repository, and it would also be a completely unnecessary dependency for users who don't need that font bundled.

I already tested stripping the unnecessary glyphs to get it down to a few MB and making such fonts downloadable under a special flag, but that didn't work for the users who tested it - https://github.com/qarmin/czkawka/issues/1987

Qianli-Ma:
That makes sense. I removed the bundled 20 MB font and changed Krokiet to use system fonts through Fontique instead.

The implementation:

- Detects installed system fonts at startup.
- Uses locale-aware Han fallback for `zh-CN` and `zh-TW`.
- Prefers common Chinese fonts such as Microsoft YaHei, Microsoft JhengHei, Noto Sans CJK, and SimSun.
- Gracefully continues when no suitable CJK font is installed.
- Leaves the normal UI font unchanged, so paths and other text retain their expected rendering.
- Logs the selected fallback font to help diagnose missing glyphs.

PR #2024, opened 2026-08-04
Added Support for .desktop in Wayland and X11 on Winit

BetoNetu:
Added support for .desktop files by:
1) Making so that de binary would be able to pass the xdg_app_id variable so the WM can correctly assign the app's icon;
2) Added a simples shellscript that creates the .desktop on ~/.local/share/applications/krokiet.desktop and puts the png on ~/.local/share/icons/...;
3) Had to add i-slint-backend-winit (1.17.0) as a dependency;

Even thought It is a "symple fix" it took me the whole day scouring forums and testing stuff until it worked. I've had a help from AI since I am not that versed in Rust, but I tested and it works (X11 and Wayland) and to not have any vulnerability since it only passes a variable throught a native slint functio

n and acess everything throught native functions. 

Also this script could help in a future deb package if they get rust up to 1.97.

Anyways happy to contibute and to help
Captura de tela de 2026-08-03 22-23-02
Captura de tela de 2026-08-03 22-22-56

### DIFF
--- a/krokiet/create-desktop-entry
+++ b/krokiet/create-desktop-entry
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+#--------------------------------------
+#Creates a Functional .desktop icon for the distros it supports by falling back into X11 because Slint don't like acessing WAYLAND_APP_ID
+#--------------------------------------
+#Variables
+BIN="${HOME}/.cargo/bin/krokiet"
+ICON_SRC="./icons/krokiet_logo_flag.png"
+APPS_DIR="${HOME}/.local/share/applications/"
+ICON_DIR="${HOME}/.local/share/icons/apps/krokiet/"
+DESKTOP_FILE="${APPS_DIR}/krokiet.desktop"
+#Functions
+handle_error() {
+    local type="$1"
+    printf "Error: %s\n" "${type}"
+    printf "Exiting...\n"
+    exit 1
+}
+check_dir() {
+    local dir="$1"
+    if [ -d "${dir}" ]; then
+        printf "%s found\n" "${dir}!"
+    else
+        printf "%s not found, trying to create...\n" "${dir}"
+        mkdir -p "${dir}" || handle_error "Failed to create ${dir}!"
+        printf "%s created\n" "${dir}"
+    fi
+}
+#Main
+if [ ! -f "${BIN}" ]; then
+    handle_error "Krokiet binary not found: ${BIN }"
+fi
+if [ ! -f "${ICON_SRC}" ]; then
+    handle_error "Source icon not found: ${ICON_SRC}"
+fi
+check_dir "$APPS_DIR"
+check_dir "$ICON_DIR"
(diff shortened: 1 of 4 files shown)


krokiet/create-desktop-entry
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+#--------------------------------------
+#Creates a Functional .desktop icon for the distros it supports by falling back into X11 because Slint don't like acessing WAYLAND_APP_ID
```
BetoNetu: Forgot to change this comment, beacause I was first only able to use it by WAYLAND_DISPLAY= krokiet

BetoNetu:
Quality check fail reason:
 rustup default 1.94.1
  rustup component add rustfmt
  rustup component add clippy
  shell: /usr/bin/bash -e {0}
- apt rust is too old, krokiet wants >= 1.97

BetoNetu:
fixes #2027

krokiet/src/main.rs
```diff
@@ -93,6 +95,9 @@ mod ui {
 pub use ui::*;
 
 fn main() {
+    let backend = Backend::new().unwrap();
+    set_platform(Box::new(backend)).unwrap();
+    slint::set_xdg_app_id("krokiet").unwrap();
```
BetoNetu, line 100: Maybe there'll be necessary to add the other Rust Window compositors as options and check them. But as a proof of concept it may work

krokiet/create-desktop-entry
BetoNetu: Already sorted, but the problem remained the same and the fix stays the same

krokiet/src/main.rs
BetoNetu: After the 12.02 update I had to manually redo this (no big deal). Also I am not aware of another way to do that other than this (tried to find something about in slint/winit docs and did not found a clue) I got his idea from an issue months ago on slint's gh

krokiet/create-desktop-entry
BetoNetu, line 1: There's probably a way to make rust call this script when installing krokiet via cargo. Like a post install in deb.

krokiet/src/main.rs
BetoNetu: This is probably a band-aid solution for most cases (since most people would not change the default rendering options) but it'll not work specifically for other backends that not winit. I do not know how/if the other ones handle xdg variables WMClass/appid/StartupWMClass. They probably should do it in their own way.

BetoNetu:
Android check seems broken as I did not change nothing there

BetoNetu:
This is probably a band-aid solution for most cases (since most people would not change the default rendering options) but it'll not work specifically for other backends that not winit. I do not know how/if the other ones handle xdg variables WMClass/appid/StartupWMClass. They probably should do it in their own way.

BetoNetu:
There are quite a few commits here because I had to work through some compiler/clippy issues in gh on web. If preferred, this can be squashed into a single commit when merging.

PR #2043, opened 2026-08-18
Fix popup OK/Close button truncation if the text is too long (#2028)

### LINKED ISSUE #2028: [Krokiet] "Move items to trash" popup has broken layout when translation is too long
**Bug Description**
When using a localization with long strings (e.g., Russian or French), the "Move items to trash" confirmation dialog does not properly display the OK and Cancel buttons — they become misaligned, truncated, or fall outside the visible area.

**Steps to reproduce:**
1. Go to Similar Images.
2. Run a scan using the Scan button.
3. Select any duplicate image by checking it.
4. Click the Trash button.

```
N/A
```

**System**

- Krokiet version: 12.0.1 gui 
- OS version: CachyOS
- Installation method: Official repositories

**Screenshot**

Image

dylanheller:
## Issue fix
`PopupTrash` had a hardcoded `popup_height` which is not enough to fit long translations. To fix it we can pass the minimum height into `PopupBase` and it will choose the maximum of `popup_min_height` and its text's preferred height.
This change was applied to all popups that use `PopupBase`.

## Popup centering
Types like `PopupTrash` have the size of the main window and contain a smaller popup window. Previously, the popup was placed in container's top-left corner and we were centering the container itself.
I placed most of the popup containers that use `PopupBase` or `PopupWindow` at (0, 0) and applied the centering to the popup inside.

Fixes #2028

### DIFF
--- a/krokiet/ui/components/popup_base.slint
+++ b/krokiet/ui/components/popup_base.slint
@@ -9,20 +9,22 @@ export component PopupBase inherits PopupWindow {
     in-out property <string> cancel_text <=> Translations.cancel_button_text;
     in-out property <bool> enabled_ok_button: true;
     in-out property <bool> show_cancel_button: true;
+    in property <length> popup_min_height: 0px;
 
     callback ok_clicked();
     callback cancel_clicked();
 
+    height: max(popup_min_height, vlayout.preferred-height);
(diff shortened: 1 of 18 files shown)


qarmin:
This also fixed positioning of popup, when resizing window

PR #2058, opened 2026-09-03
Krokiet: resizable bottom panel and button to exclude all selected items

Tamalero:
Two small, independent Krokiet UI improvements, one commit each. Happy to split them into separate PRs if you prefer.

## 1. Resizable bottom panel (`9f29468`)

The included/excluded paths panel is pinned by a hardcoded `min-height: 150px` in `bottom_panel.slint`, so it always occupies roughly the bottom fifth of the window and cannot be resized. On a large results list that space is either wasted or too small to review many paths.

This adds a `HorizontalSplitter` component (`ui/components/splitter.slint`): a 7px drag bar with an `ns-resize` cursor, placed between the results area and the action buttons. The panel height now comes from `Settings.bottom_panel_height` and is persisted in `BasicSettings`, so it survives a restart. The old 150px minimum becomes an 80px floor.

Two implementation notes:

- The splitter caches the panel height on pointer-down and applies the total drag delta, so the bar tracks the cursor 1:1. The existing column-resize handle in `selectable_tree_view.slint` uses `+=` on every `moved` event, which accumulates the delta and makes the drag accelerate.
- The clamp that keeps the panel from eating the results list on window shrink lives in a `changed height` handler, not in the panel's `height` binding. A declarative `min(..., root.height - ...)` there creates a binding loop (`bottom-panel-max-height -> bottom-panel.height -> layoutinfo-v -> root.layoutinfo-v -> ...`) which Slint 1.17 reports as deprecated and "may cause panic at runtime".

## 2. Exclude all selected items (`c046a1f`)

Excluding items is currently one at a time through the row context menu. After a scan that surfaces many unwanted matches, that is a lot of clicking.

This adds an `Exclude` button next to Hardlink/Softlink that adds every checked item to the excluded paths list in one action, behind the same `PopupActionConfirm` the link actions use, showing the number of selected items. The rows are then removed from the results, groups left orphaned are cleaned up via `remove_single_items_in_groups`, and the selection counter is reset.

Full file paths land in `Directories::excluded_files` (`process_paths` routes on `is_file()`), so the next scan skips them. This reuses the existing `add_excluded_paths`, which already deduplicates and sorts.

The button is enabled only when something is checked, and it is available on every tool tab.

## Testing

- `cargo check -p krokiet`: clean, no warnings
- `cargo test -p krokiet`: passing, including two new unit tests for the checked-path collection
- `slint-viewer --check` on `main_window.slint`: clean, no binding-loop warnings
- Verified the first commit builds standalone as well, since both commits touch `main_window.slint`
- Panel resize verified by rendering the window at several stored heights and by manual dragging; the value round-trips through the config file
- `misc/find_unused_callbacks.py`, `find_unused_slint_translations.py` and `find_unused_fluent_translations.py` all report nothing unused for the new callback and translation keys

Only the English `.ftl` file is touched, per the Crowdin note in `AGENTS.md`.


Tamalero:
Claude + CachyOS tested.  I apologize if I busted the etiquette for merging in this project.

PR #2059, opened 2026-09-03
Reject perceptual hashes that encode no spatial content

Tamalero:
Similar Images can group completely unrelated pictures as `Original`, and no setting avoids it: the images collide at Hamming distance 0, so raising the hash size or lowering Max difference does not help.

## What happens

An image made of one light region and one dark region (a sketch beside its finished version, a light and a dark colourway of the same pattern, and so on) hashes to a bit matrix whose rows are all identical.

Blockhash and Mean set each block's bit by comparing it against the median of all blocks. On a strongly bimodal image every block on the light side lands above the median and every block on the dark side below it, so the picture's own detail never crosses the threshold and contributes no bits at all. What is left encodes only the position of the split, and any two images sharing that split are identical to the hasher.

Measured on 7 such files reported by a user, hashed exactly as `SimilarImages` does (Blockhash, Lanczos3):

| hash size | max pairwise distance between the 7 unrelated images |
|---|---|
| 16 | 0 |
| 32 | 2 |
| 64 | 80 |

At 16 and 32 they are indistinguishable. At 64 they separate on their own, which is why the fix below does not need to catch them all there.

`is_hash_valid` did not stop this. It rejects a hash that is entirely `0x00` or entirely `0xFF`, but a split hash is exactly 50% ones and passes.

## The fix

Reject a hash whose bit matrix has fewer than 4 distinct rows or fewer than 4 distinct columns. A hash that degenerate carries no usable content, in the same spirit as the existing all-black / all-white check.

Two deliberate limits:

- **Only from hash size 16 up.** An 8x8 matrix has too few lines to tell a degenerate hash from a legitimately simple picture; applying the rule there rejected 0.2 to 1.7 percent of a real sample, so it is skipped.
- **Only for square layouts.** `DoubleGradient` does not produce a `hash_size x hash_size` matrix, so the check cannot interpret its bits and passes it through untouched.

## Evidence

4000 images from a personal photo and artwork library, against the 7 reported files:

| | reported files score | library minimum | caught | false rejects |
|---|---|---|---|---|
| Blockhash 16 | 1 | 7 | 7/7 | 0 / 4000 |
| Blockhash 32 | 1 to 3 | 12 | 7/7 | 0 / 4000 |
| Blockhash 64 | 1 to 18 | 25 | 3/7 | 0 / 4000 |

The library minimum stays far above the threshold of 4 at every applicable size, and nothing in the sample was rejected.

I first tried a bit-transition ratio instead. It looked convincing on 596 images but degraded to a 1.3x margin at 3928 with false rejections that grew as the sample grew, so it was dropped in favour of the structural measure above, which separates by roughly an order of magnitude.

The gradient algorithms are unaffected by the underlying problem: they compare adjacent pixels rather than a global median, so the same 7 files score 18 to 32 and never collided. Switching the hash algorithm is a usable workaround for anyone hitting this today.

Worth noting: a perfectly smooth synthetic gradient is itself close to structureless and would be rejected. That is arguably correct, since its hash carries almost nothing, but it is a behaviour change and I would rather flag it than hide it.

## Tests

Five new tests in `similar_images/tests.rs`: unrelated split-tone images collide *and* are rejected at 16/32/64; detailed images are kept; the check is skipped at hash size 8; uniform hashes are still rejected; a non-square `DoubleGradient` layout is passed through.

## Checks run locally

- `cargo fmt --all -- --check` clean
- `cargo clippy --all-targets --all-features -- -D warnings` and `cargo clippy --all-targets -- -D warnings` both clean
- `cargo test -p czkawka_core` - 241 passed, 0 failed

Local toolchain is 1.92 rather than the pinned 1.94.1, so clippy there may still see something mine did not.

Note this changes hashing behaviour but not the hash values themselves, so existing caches stay valid.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,8 @@
+## Version 12.0.2 - 03.09.2026r
+
+### Core
+- Fixed unrelated images that consist of one light and one dark region being grouped as identical - such images produced a perceptual hash encoding only the position of the split, so they collided at distance 0 regardless of hash size - [#2059](https://github.com/qarmin/czkawka/pull/2059)
+
 ## Version 12.0.1 - 29.07.2026r
 
 ### Core
--- a/czkawka_core/src/tools/similar_images/core.rs
+++ b/czkawka_core/src/tools/similar_images/core.rs
@@ -21,7 +21,10 @@ use crate::common::progress_data::{ProgressData, SimilarImagesStage, ToolStage};
 use crate::common::progress_stop_handler::{check_if_stop_received, prepare_thread_handler_common};
 use crate::common::tool_data::{CommonData, CommonToolData};
 use crate::flc;
-use crate::tools::similar_images::{GeometricInvariance, Hamming, ImHash, ImagesEntry, SIMILAR_VALUES, SimilarImages, SimilarImagesParameters, SimilarityPreset};
+use crate::tools::similar_images::{
+    GeometricInvariance, Hamming, ImHash, ImagesEntry, MIN_DISTINCT_HASH_LINES, MIN_HASH_SIZE_FOR_STRUCTURE_CHECK, SIMILAR_VALUES, SimilarImages, SimilarImagesParameters,
+    SimilarityPreset,
+};
 
 impl SimilarImages {
     pub fn new(params: SimilarImagesParameters) -> Self {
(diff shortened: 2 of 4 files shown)


Tamalero:
This happened in images that had some interesting dynamic.
All images affected (were grouped together despite being completely different) were split perfectly in half where one side had a dark version and the other a brighter version (or opposite). aka some sort of large variation of brightness.

Example of one of the affected files: 
66126400_loupgarou_1787498198 loupgarou_202608_uv_diwali_compo

